PR #8, opened 2020-09-15
Sub Entities. Yes please!

### LINKED ISSUE #9: Declaring some components with customHypermediaElement results in nothing rendered
There's a problem with rendering when attempting to declare `d2l-activity-editor-footer` with `customHypermediaElement`. Nothing is rendered. When inspecting, one sees `<!--{{lit-18816010524707827}}-->` where the rendered footer should be.
  - To reproduce, change Line 31 in `d2l-activity-editor-footer.js` to `customHypermediaElement('d2l-activity-editor-footer', ActivityEditorFooter, 'd2l-activity-editor-footer');`
  - Add import statement `import { customHypermediaElement } from '../../../framework/hypermedia-components.js';`
  - Visit the dev site and observe non-rendered footer:

### LINKED ISSUE #10: Activity name rendered outside of wrapping h1 tag in activity editor
`activity-name` is rendered outside of the wrapping `h1` tag in `d2l-activity-editor-header`

### LINKED ISSUE #11: Bad siren component
Viewing the polaris dev site results in an unknown error: `hypermedia-lit-mixin.js:55 Error: Bad siren component`

Could be due to a bad `throw`

m6jones:
This basically gives us the ability to pull a list of sub entities.

image

Each one is the different course items for the learning path. Once I get my split mixin for list items in we can add a course list.

Also, fixed some other issues. 
- Fixes #9 
- Fixes #10
- Fixes #11

I also added the ability to add `method` to give us more options. I ended up not needing this change yet after talking to Eric. But could be super useful. I would want the method to also send the raw siren object for more options but lets do that if we need it.

### DIFF
--- a/framework/hypermedia-components.js
+++ b/framework/hypermedia-components.js
@@ -5,6 +5,7 @@ import { observableTypes } from '../state/HypermediaState.js';
 import { until } from 'lit-html/directives/until.js';
 
 export function customHypermediaElement(tag, elementClass, pseudoTag, hypermediaClasses, options) {
+	pseudoTag = pseudoTag || tag;
 	const components = componentStoreFactory(pseudoTag);
 	components.register(tag, hypermediaClasses);
 	customElements.define(tag, elementClass, options);
@@ -46,8 +47,14 @@ export function html(strings, ...values) {
 		});
 
 		currentCollection.strings.push(currentString);
-		currentValue && currentCollection.values.push(currentValue);
+		currentCollection.values.push(currentValue);
 	}
+
+	//todo: this solves the token ifDefined issue. But really? Shouldn't there be a real way to fix it? I think so.
+	while (stringCollections[0].strings.length <= stringCollections[0].values.length && stringCollections[0].values.length !== 0) {
+		stringCollections[0].values.pop();
+	}
+
 	return new TemplateResult(stringCollections[0].strings, stringCollections[0].values, 'html', defaultTemplateProcessor);
 }
 
--- a/framework/hypermedia-lit-mixin.js
+++ b/framework/hypermedia-lit-mixin.js
@@ -26,12 +26,11 @@ export const HypermediaLitMixin = superclass => class extends superclass {
 	}
 
 	updated(changedProperties) {
-		if ((changedProperties.has('href') || changedProperties.has('token')) &&
-			this.href && this.token) {
-			dispose(this._state);
+		if ((changedProperties.has('href') || changedProperties.has('token')) && this.href && this.token) {
+			dispose(this._state, this);
 			this._makeState();
 		}
-		super.updated();
+		super.updated(changedProperties);
 	}
 
 	connectedCallback() {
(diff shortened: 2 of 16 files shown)


framework/hypermedia-components.js
```diff
@@ -46,8 +47,14 @@ export function html(strings, ...values) {
 		});
 
 		currentCollection.strings.push(currentString);
-		currentValue && currentCollection.values.push(currentValue);
+		currentCollection.values.push(currentValue);
 	}
+
+	//todo: this solves the token ifDefined issue. But really? Shouldn't there be a real way to fix it? I think so.
```
zinabat, line 53: 🤔 I'm gonna say I can't think of a better way to do this and it's a big issue right now, so this is fine

framework/hypermedia-lit-mixin.js
```diff
@@ -26,12 +26,11 @@ export const HypermediaLitMixin = superclass => class extends superclass {
...
 	updated(changedProperties) {
-		if ((changedProperties.has('href') || changedProperties.has('token')) &&
-			this.href && this.token) {
-			dispose(this._state);
+		if ((changedProperties.has('href') || changedProperties.has('token')) && this.href && this.token) {
+			dispose(this._state, this);
 			this._makeState();
 		}
-		super.updated();
+		super.updated(changedProperties);
```
zinabat, line 33: Good catch

state/HypermediaState.js
```diff
@@ -1,13 +1,7 @@
...
 import { refreshToken } from './token.js';
 
-export const observableTypes = Object.freeze({
-	property: 1,
-	link: 2,
-	classes: 3,
-	subEntities: 4,
-	entity: 5
-});
+export const observableTypes = ot;
```
zinabat, line 4: why do we need to export this in two places?

zinabat (commented):
I just did a first pass, but like the way this is looking.

framework/hypermedia-components.js
m6jones: Now if you add `ifDefined` things will break.

state/HypermediaState.js
m6jones: can we export only the one thing? 
This makes it so the user only needs to import HypermediaState and not an inner file.

state/sirenComponents/sirenComponentFactory.js
```diff
@@ -2,6 +2,7 @@ import { SirenClasses } from './SirenClasses.js';
 import { SirenEntity } from './SirenEntity.js';
 import { SirenLink } from './SirenLink.js';
 import { SirenProperty } from './SirenProperty.js';
+import { SirenSubEntities } from './SirenSubEntities.js';
 import { SirenSubEntity } from './SirenSubEntity.js';
```
zinabat, line 6: The difference here is for sub-entities that are plural in nature, such as for collections, yes?

zinabat (approved):
Just a couple clarifications but otherwise good to go

PR #13, opened 2020-09-17
Ccisse/collection

ciss1995:
Context:
US120224 - Refactor activity-editor-main-collection to use LP layout and be used in activity-editor. 

Screenshot 2020-09-16 131854
Screenshot 2020-09-17 112816


components/activity/editor/d2l-activity-editor.js
```diff
@@ -59,7 +59,7 @@ class ActivityEditor extends LitElement {
 		return html`
 			<div class="d2l-activity-editor-template-default">
 				<d2l-activity-editor-header href="${this.href}" .token="${this.token}"></d2l-activity-editor-header>
-				<d2l-activity-editor-main href="${this.href}" .token="${this.token}"></d2l-activity-editor-main>
+				<d2l-activity-editor-main-collection href="${this.href}" .token="${this.token}"></d2l-activity-editor-main-collection>
```
zinabat, line 62: Can we put this back to be just `main`?

components/activity/editor/d2l-activity-collection-editor.js
```diff
@@ -15,4 +15,4 @@ class ActivityEditorMain extends HypermediaLitMixin(LitElement) {
 	}
 }
 
-customHypermediaElement('d2l-activity-editor-main', ActivityEditorMain, 'd2l-activity-editor-main');
+customHypermediaElement('d2l-activity-ecollection-editor', ActivityCollectionEditorMain, 'd2l-activity-ecollection-editor-main');
```
zinabat, line 18: For this file, since we don't have a default `main` you can just add an import statement for the `acitivity-editor-main-collection`. I'd also rename it to what it was before.

components/activity/editor/d2l-activity-editor-main-collection.js
```diff
@@ -0,0 +1,73 @@
...
+					<d2l-list>
+						${this.items.map(href => html`<d2l-list-item>${href}</d2l-list-item>`)}
+					</d2l-list>
+				</div>
+			</div>
+		`;
+	}
+}
+
+customHypermediaElement('d2l-activity-editor-main-collection', ActivityEditorMainCollection, 'd2l-activity-editor-main-collection');
```
zinabat, line 73: Likewise with other comments, this should extend the tag `d2l-activity-editor-main` and resolve on learning-path or activity-collection
```suggestion
customHypermediaElement('d2l-activity-editor-main-collection', ActivityEditorMainCollection, 'd2l-activity-editor-main', [['activity-collection'], ['learning-path']]);

```

components/activity/editor/d2l-activity-editor.js
```diff
@@ -72,7 +72,7 @@ class ActivityEditor extends LitElement {
 			<d2l-template-primary-secondary>
 				<slot name="editor-nav" slot="header"></slot>
 				<d2l-activity-editor-header slot="primary" href="${this.href}" .token="${this.token}"></d2l-activity-editor-header>
-				<d2l-activity-editor-main slot="primary" href="${this.href}" .token="${this.token}"></d2l-activity-editor-main>
+				<d2l-activity-editor-main-collection slot="primary" href="${this.href}" .token="${this.token}"></d2l-activity-editor-main-collection>
```
zinabat, line 75: Same thing here. We want this as just `main` so that for instance `d2l-activity-editor-main-assignment` will get resolved when we add it.

components/activity/editor/d2l-activity-editor.js
ciss1995: Yeah totally!

components/activity/editor/d2l-activity-collection-editor.js
ciss1995: Shouldn't we just remove this file actually? and call `d2l-activity-editor-main-collection.js` `d2l-activity-editor-main.js`?

components/activity/editor/d2l-activity-editor-main-collection.js
ciss1995: Oh yeah forgot about that!

components/activity/editor/d2l-activity-collection-editor.js
zinabat: I wouldn't since the collection version should only resolve when there is an activity collection or learning path class present. For now, we can leave it as no default.

zinabat (approved):

PR #14, opened 2020-09-17
Added readmes

zinabat:
## Context
[US120235](https://rally1.rallydev.com/#/detail/userstory/424740820104?fdp=true)

- Adds readme files for all current components
- Adds a readme file for creating a new component
  - Missing action information since it's not completed yet
- Updates the index readme to explain what it does

When reviewing, I recommend viewing the actual branch to see the pretty versions of the readme files.

### DIFF
--- a/creating-new-components.md
+++ b/creating-new-components.md
@@ -31,10 +31,10 @@ customHypermediaElement(
 
 ### Determining When a Tag Will Load
 
-Base tags like `d2l-activity-editor-main` resolve to other components based on the hypermedia classes on the entity. We can customize this logic using basic syntax.
+Base tags like `d2l-activity-editor-main` resolve to other components based on the hypermedia classes on the entity. We can customize this logic using the following syntax:
 
-Add classes to an array to `AND` them.
-Add more arrays of classes to `OR` them.
+- Add classes to an array to `AND` them.
+- Add more arrays of classes to `OR` them.
 
 |Argument|Resolve when classes include|
 |-|-|
--- a/creating-new-components.md
+++ b/creating-new-components.md
@@ -33,17 +33,28 @@ customHypermediaElement(
 
 Base tags like `d2l-activity-editor-main` resolve to other components based on the hypermedia classes on the entity. We can customize this logic using the following syntax:
 
-- Add classes to an array to `AND` them.
+- Add classes to an inner array to `AND` them.
 - Add more arrays of classes to `OR` them.
+
+A single class does not need to be wrapped in an array.
 
 |Argument|Resolve when classes include|
 |-|-|
 |`[['learning-path']]`| `learning-path`|
+|`'learning-path'`| `learning-path`|
 |`[['activity-collection'], ['learning-path']]`| `activity-collection` OR `learning-path`|
 |`[['activity-usage', 'course-offering']]`| `activity-usage` AND `course-offering`|
 |`[['activity-usage', 'course-offering'], ['quiz']]`| (`activity-usage` AND `course-offering`) OR `quiz`|
+|`[['activity-usage', 'course-offering'], 'quiz']`| (`activity-usage` AND `course-offering`) OR `quiz`|
 
 Tag resolution will always use the most specific logic available; a component defined with a longer `AND` statement will be preferred over a component defined with a shorter statement.
+
+**Example:** Two components are defined as follows:
+
+- `component1` defined with class argument `[['foo', 'bar'], ['cherry', 'coke']]`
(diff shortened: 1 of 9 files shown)


creating-new-components.md
```diff
@@ -0,0 +1,100 @@
...
+Add more arrays of classes to `OR` them.
+
+|Argument|Resolve when classes include|
+|-|-|
+|`[['learning-path']]`| `learning-path`|
+|`[['activity-collection'], ['learning-path']]`| `activity-collection` OR `learning-path`|
+|`[['activity-usage', 'course-offering']]`| `activity-usage` AND `course-offering`|
+|`[['activity-usage', 'course-offering'], ['quiz']]`| (`activity-usage` AND `course-offering`) OR `quiz`|
+
+Tag resolution will always use the most specific logic available.
```
tenbradley, line 46: "specific logic"? Does this mean if component1 has tags `foo` and `bar`, and component2 has tags `foo`, `bar`, and `baz`, that it resolves to component2?
If so, is specific counted as the longest 'and' statement that is true, or the logic statement with the most terms?
Either way, I think this should be more detailed.

creating-new-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Registering
+
+Most use cases will be shareable and the new components should be added to this repository. If your tool has custom logic that is not reusable, the component can live in a different repository.
+
+The new component **needs to be imported** in the base component. In this case, in `d2l-activity-editor-main.js`:
```
tenbradley, line 52: Is this a "needs to be imported or it will not work", or "needs to be imported or we will hurt you" statement? I assume the former. (I don't think this really **needs** to be clarified, but it's a question one could ask.)

creating-new-components.md
```diff
@@ -0,0 +1,100 @@
...
+  [['some-class']] // the entity classes to determine when this component is used
+);
+```
+
+### Determining When a Tag Will Load
+
+Base tags like `d2l-activity-editor-main` resolve to other components based on the hypermedia classes on the entity. We can customize this logic using basic syntax.
+
+Add classes to an array to `AND` them.
+Add more arrays of classes to `OR` them.
```
AKobets, line 37: Why we need OR, is it explained ?

zinabat: One use case is learning paths, where most learning-path specific components should match on either `learning-path` OR `activity-collection`

creating-new-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+|Argument|Resolve when classes include|
+|-|-|
+|`[['learning-path']]`| `learning-path`|
+|`[['activity-collection'], ['learning-path']]`| `activity-collection` OR `learning-path`|
+|`[['activity-usage', 'course-offering']]`| `activity-usage` AND `course-offering`|
+|`[['activity-usage', 'course-offering'], ['quiz']]`| (`activity-usage` AND `course-offering`) OR `quiz`|
+
```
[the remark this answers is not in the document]
zinabat: We can probably make that work.

creating-new-components.md
zinabat: That's a good question that I'll need to investigate the answer to.

creating-new-components.md
zinabat: The former. Ideally this is a step we will automate in the future. _Not_ importing the component will result in the next matching component to be resolved. In most cases, it will just resolve to the default.

creating-new-components.md
m6jones: Yup if it has 3 classes it will resolve the 3 class one of 2 class one.

creating-new-components.md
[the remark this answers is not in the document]
m6jones: I honestly think I made this a feature already. I'm all about overloaded methods.

creating-new-components.md
m6jones: I could see it being used if a component can show a name and/or description.

creating-new-components.md
tenbradley: Do things break if there's two statements with the same number of classes? Is that even possible?

creating-new-components.md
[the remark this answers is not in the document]
zinabat: Oh shit, you're right; this works:
`customHypermediaElement('d2l-activity-editor-main-assignment', ActivityEditorMainAssignment, 'd2l-activity-editor-main', 'assignment-activity');`

creating-new-components.md
m6jones: Grabs the first one. Ie, randomly selects one in a consistent way.

tenbradley (approved):

PR #21, opened 2020-09-26
SelflessEntities they work!

m6jones:
image

### DIFF
--- a/components/activity/editor/d2l-activity-editor-type-assignment.js
+++ b/components/activity/editor/d2l-activity-editor-type-assignment.js
@@ -21,7 +21,7 @@ class ActivityEditorTypeAssignment extends RtlMixin(HypermediaLitMixin(LitElemen
 		return {
 			// todo: we don't have selfless entities, so we're just grabbing the raw thing for now
 			entity: { type: Object, observable: observableTypes.entity },
-			groupCategoryName: { type: String, observable: observableTypes.property, id: "groupName",  route:[{observable: observableTypes.subEntity, rel: rels.folderType}]},
+			groupCategoryName: { type: String, observable: observableTypes.property, id: 'groupName',  route:[{observable: observableTypes.subEntity, rel: rels.folderType}]},
 			_folderTypeClasses: { type: Object, observable: observableTypes.classes, route:[{observable: observableTypes.subEntity, rel: rels.folderType}]}
 		};
 	}
--- a/components/activity/editor/d2l-activity-editor-type-assignment.js
+++ b/components/activity/editor/d2l-activity-editor-type-assignment.js
@@ -20,7 +20,9 @@ class ActivityEditorTypeAssignment extends RtlMixin(HypermediaLitMixin(LitElemen
 	static get properties() {
 		return {
 			// todo: we don't have selfless entities, so we're just grabbing the raw thing for now
-			entity: { type: Object, observable: observableTypes.entity }
+			entity: { type: Object, observable: observableTypes.entity },
+			groupCategoryName: { type: String, observable: observableTypes.property, id: "groupName",  route:[{observable: observableTypes.subEntity, rel: rels.folderType}]},
+			_folderTypeClasses: { type: Object, observable: observableTypes.classes, route:[{observable: observableTypes.subEntity, rel: rels.folderType}]}
 		};
 	}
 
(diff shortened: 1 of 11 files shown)


zinabat:
Omg omg omg it's sooooo simple <3

Will check this tomorrow

components/activity/editor/d2l-activity-editor-type-assignment.js
```diff
@@ -20,7 +20,9 @@ class ActivityEditorTypeAssignment extends RtlMixin(HypermediaLitMixin(LitElemen
 	static get properties() {
 		return {
 			// todo: we don't have selfless entities, so we're just grabbing the raw thing for now
-			entity: { type: Object, observable: observableTypes.entity }
+			entity: { type: Object, observable: observableTypes.entity },
+			groupCategoryName: { type: String, observable: observableTypes.property, id: 'groupName',  route:[{observable: observableTypes.subEntity, rel: rels.folderType}]},
+			_folderTypeClasses: { type: Object, observable: observableTypes.classes, route:[{observable: observableTypes.subEntity, rel: rels.folderType}]}
```
zinabat, line 25: This seems to get set to the classes on the entity instead of the sub entity.
I'm getting `["named-entity", "assignment"]` instead of the expected `["folder-type", "individual", "no-group-type"]`

components/activity/editor/d2l-activity-editor-type-assignment.js
```diff
@@ -20,7 +20,9 @@ class ActivityEditorTypeAssignment extends RtlMixin(HypermediaLitMixin(LitElemen
 	static get properties() {
 		return {
 			// todo: we don't have selfless entities, so we're just grabbing the raw thing for now
-			entity: { type: Object, observable: observableTypes.entity }
+			entity: { type: Object, observable: observableTypes.entity },
```
zinabat, line 23: These lines can be removed now.

components/activity/editor/d2l-activity-editor-type-assignment.js
```diff
@@ -63,22 +65,7 @@ class ActivityEditorTypeAssignment extends RtlMixin(HypermediaLitMixin(LitElemen
...
-	}
-
-	// this should also be made moot by selfless entities
-	get groupCategoryName() {
-		if (this.entity) {
-			const folderType = this.entity.getSubEntityByRel(rels.folderType);
-			return folderType && folderType.properties && folderType.groupName;
-		}
-		return null;
+		return this._folderTypeClasses?.some((item) => item === assignmentTypes.individual);
```
zinabat, line 68: Very excited to have this be one line <3

components/activity/editor/d2l-activity-editor-type-assignment.js
m6jones: Well, something is broken :(

Oh, this leads to an issue right now where if you typo your route you are screwed with no indication anything went wrong. We need a solution for that.

components/activity/editor/d2l-activity-editor-type-assignment.js
m6jones: I am thinking we also have a `observableType` class that value is a boolean.

framework/hypermedia-lit-mixin.js
```diff
@@ -59,3 +63,23 @@ export const HypermediaLitMixin = superclass => class extends superclass {
 		return this[action] && this[action].has;
 	}
 };
+
+function deepCopy(inObject) {
+
+	if (typeof inObject !== 'object' || inObject === null) {
+		return inObject; // Return the value if inObject is not an object
```
AKobets, line 70: should it return something else, like null or undefined if it is not an object?

framework/hypermedia-lit-mixin.js
```diff
@@ -59,3 +63,23 @@ export const HypermediaLitMixin = superclass => class extends superclass {
 		return this[action] && this[action].has;
 	}
 };
+
+function deepCopy(inObject) {
```
AKobets, line 67: is there already reusable solution for deep copy?

AKobets: would https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign work instead of creating custom function?

dbatiste: The `Object.assign` helper does a shallow copy - that is, it will copy the primitives, and it will copy references to the reference types, so the copy ends up pointing to the same object.

I agree we don't want this sort of thing being duplicated, and it seems so generic that it belongs somewhere that it could be reused if needed.  Ideally it would have a bunch of tests associated with it, that covers a variety of types, including the primitives, arrays, objects, dates, null/undefined, maps, etc.

components/activity/editor/d2l-activity-editor-type-assignment.js
AKobets: > Well, something is broken :(
> 
> Oh, this leads to an issue right now where if you typo your route you are screwed with no indication anything went wrong. We need a solution for that.

may be throw exception

framework/hypermedia-lit-mixin.js
zinabat: Maybe put this method in a helpers file. I do appreciate that we're not using `lodash` here because.. y'know.

zinabat (approved):
Tested!

PR #23, opened 2020-09-29
Preventing extrenuous undefines

tenbradley:


AKobets:
@m6jones  is it correct syntax here https://github.com/BrightspaceHypermediaComponents/polaris/pull/23/files#diff-093aad5546909e033e88800812dcb489L14
`export function html(strings, ...values) {`
Here example of trying to undestand what it does.

lets have `function x(strings, ...values){console.log(values)}  `

then `x(['1','2'],[3,4]);` prints `[ [ 3, 4 ] ]` 
and `x(['1','2'],3,4);` prints `[ 3, 4 ]`
hml is called in render as `html(strings, values)`, so inside html function the `values` will be `[values]`

confusing.

Issue #17, opened 2020-09-21
Classes named with the component tag name included are replaced with the resolved tag name


zinabat:
`<d2l-activity-editor-main class="d2l-activity-editor-main" href="foo" token="bar">`

If the above tag will resolve to `d2l-activity-editor-main-assignment`, then the class also resolves to `d2l-activity-editor-main-assignment`. This is probably counter-intuitive for developers and not desired behaviour.

**Workaround:** 
Target the element with attribute prefixing:
`[class^="d2l-activity-editor-main"]` will match anything beginning with `d2l-activity-editor-main` in its class name

PR #24, opened 2020-09-30
More specific pseudotag replace

tenbradley:
Kind of a hack.
Matching "<"+pseudotag"+" " or "</"+pseudotag"+" " for replacing pseudotags.


framework/HypermediaResult.js
```diff
@@ -0,0 +1,145 @@
...
+		}
+		strings.pop();
+		if (strings.length !== 0) {
+			mainStrings.push(`</${tag}>`);
+			mainValues.push(html(strings, values));
+		} else {
+			mainStrings[mainStrings.length - 1] += `</${tag}>`;
+		}
+
+		//dispose(state);
```
m6jones, line 142: You can remove this

m6jones (approved):

framework/HypermediaResult.js
```diff
@@ -0,0 +1,145 @@
...
+export class HypermediaResult extends TemplateResult {
+
+	constructor(strings, values, type, processor) {
+		super(strings, values, type, processor);
+		this.processing();
+	}
+
+	/*
+	* Function called on HypermediaResult object to get the associated html
+	* Should process the psuedocode if this is the first call to getHTML
```
zinabat, line 28:
```suggestion
	* Should process the base tag if this is the first call to getHTML
```
I think you meant this?

framework/HypermediaResult.js
```diff
@@ -0,0 +1,145 @@
...
+	* otherwise, retrieves previous processing
+	*/
+	getHTML() {
+		return super.getHTML();
+	}
+
+	processing() {
```
zinabat, line 35: I would still love to see more comments throughout this function

framework/HypermediaResult.js
```diff
@@ -0,0 +1,144 @@
...
+	render(state, components, resources, pseudoTag, strings, values) {
+		const tag = components.componentTag(resources.classes);
+		if (!tag) return null;
+		strings = [...strings];
+		const mainStrings = [], mainValues = [];
+		while (strings.length > 0) {
+			let string = strings.shift();
+			if (!string) break;
+			string = string.replace(`<${pseudoTag} `, `<${tag} `);
+			string = string.replace(`</${pseudoTag} `, `</${tag} `);
```
zinabat, line 120: I assume this fixes #17

framework/hypermedia-components.js
```diff
@@ -1,121 +1,10 @@
...
 
 export function customHypermediaElement(tag, elementClass, pseudoTag, hypermediaClasses, options) {
-	pseudoTag = pseudoTag || tag;
-	const components = componentStoreFactory(pseudoTag);
-	components.register(tag, hypermediaClasses);
-	customElements.define(tag, elementClass, options);
+	custom(tag, elementClass, pseudoTag, hypermediaClasses, options);
 }
 
 export function html(strings, ...values) {
```
zinabat, line 8: It's so much clearer what's happening here. 🎉

framework/HypermediaResult.js
tenbradley: That is the intent. Not sure if there is a better way.

framework/HypermediaResult.js
tenbradley: I agree. I don't think this function will change much in trying to improve everything, so it is likely a good idea to comment it now.
I can work on this.

PR #26, opened 2020-10-05
save and cancel

m6jones:


components/activity/editor/d2l-activity-editor-footer.js
```diff
@@ -24,13 +25,21 @@ class ActivityEditorFooter extends HypermediaLitMixin(LitElement) {
 	render() {
 		return html`
 			<div>
-				<d2l-button primary>Save and Close</d2l-button>
-				<d2l-button>Cancel</d2l-button>
+				<d2l-button primary @click=${this._push}>Save and Close</d2l-button>
+				<d2l-button @click=${this._reset}>Cancel</d2l-button>
```
zinabat, line 29: nit: functions should be encapsulated with double quotes
nit: can we rename the handlers to `_onSaveClick` etc.?

components/activity/editor/d2l-activity-editor-header.js
```diff
@@ -1,14 +1,19 @@
...
+	specialization: 'https://api.brightspace.com/rels/specialization'
+});
 
 class ActivityEditorHeader extends HypermediaLitMixin(LitElement) {
 
 	static get properties() {
 		return {
-			subTitle: { type: String, attribute: 'sub-title' }
+			subTitle: { type: String, attribute: 'sub-title' },
+			_specializationHref: { type: String, observable: observableTypes.link, rel: rels.specialization, prime: true},
```
zinabat, line 16: Remind me what `prime` does again for docs purposes?

framework/HypermediaResult.js
```diff
@@ -206,6 +205,6 @@ export class HypermediaResult extends TemplateResult {
 
 		// return all of this information as a HypermediaResult to be stored as a value for a different Result
 
-		return new HypermediaResult(mainStrings, mainValues, 'html', defaultTemplateProcessor);
+		return new TemplateResult(mainStrings, mainValues, 'html', defaultTemplateProcessor);
```
zinabat, line 208: Why did this change back to `TemplateResult`?

state/sirenComponents/SirenAction.js
```diff
@@ -9,7 +9,7 @@ export class SirenAction {
 
 	constructor({id: name, token, state}) {
 		this._components = new Component();
-		this._action = { has: false, perform: () => undefined, update: () => undefined };
+		this._action = { has: false, commit: () => undefined };
```
zinabat, line 12: 🎉

state/HypermediaState.js
```diff
@@ -87,6 +87,30 @@ export class HypermediaState {
 		});
 	}
 
+	push() {
+		this._childStates().forEach(childState => childState.push());
+		const actions = this._getMap(this._decodedEntity, observableTypes.action);
+		actions.forEach(action => action.push());
+	}
```
zinabat, line 94: When I tested the save button on learning paths, I noticed that the action appears to be sent twice. Is this potentially because of the calls to `childState.push()`?

framework/HypermediaResult.js
m6jones: infinite loop bug. This is setting the normal template result. Hmm, maybe a better fix is required @tenbradley

components/activity/editor/d2l-activity-editor-header.js
m6jones: prime pre fetches. It primes the cache. Side effect is that right now it is the only way to connect entities... :D

state/HypermediaState.js
m6jones: I forgot to clean up :O

framework/HypermediaResult.js
tenbradley: Hmm, infinite loops are no good. :( I'll look into this more.

m6jones: Discussed offline. Ten and Maya will talk about this more not part of this PR.

zinabat:
Moved to `foundation-engine` repository

PR #43, opened 2020-11-30
Fixing demo path to d2l-activity-editor.js file

CodeBaboon:
I noticed that the demo did not appear to be loading the activity editor, determined it was due to a path error. This should fix it up.


tenbradley (approved):
Good catch. We needed to rename the component and this needed to be changed to match.

PR #46, opened 2020-12-03
Add activity description editor component

lloydwaddell:
```Context```
https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F378718811140

Created d2l-activity-description-editor component which adds a textarea that can be save / cancelled. Clicking save updates the description, while cancelling reverts it back to the previous state.

```Quality```
manual testing to ensure cancel / save works as expected 
Textarea should also appear larger when on smaller screens as per images in the rally story.

### DIFF
--- a/components/activity/description/d2l-activity-description-editor.js
+++ b/components/activity/description/d2l-activity-description-editor.js
@@ -14,8 +14,7 @@ class ActivityDescriptionEditor extends HypermediaStateMixin(LitElement) {
 
 	static get properties() {
 		return {
-			description: { type: String, observable: observableTypes.property, route: [{observable: observableTypes.link, rel: rels.specialization}],
-				hasChanged: () => true},
+			description: { type: String, observable: observableTypes.property, route: [{observable: observableTypes.link, rel: rels.specialization}]},
 			updateDescription: { type: Object, observable: observableTypes.action, name: 'update-description',
 				route: [{observable: observableTypes.link, rel: rels.specialization}]
 			}
--- a/components/activity/description/d2l-activity-description-editor.js
+++ b/components/activity/description/d2l-activity-description-editor.js
@@ -11,9 +11,11 @@ const rels = Object.freeze({
 });
 
 class ActivityDescriptionEditor extends HypermediaStateMixin(LitElement) {
+
 	static get properties() {
 		return {
-			description: { type: String, observable: observableTypes.property, route: [{observable: observableTypes.link, rel: rels.specialization}] },
+			description: { type: String, observable: observableTypes.property, route: [{observable: observableTypes.link, rel: rels.specialization}],
+				hasChanged: () => true},
 			updateDescription: { type: Object, observable: observableTypes.action, name: 'update-description',
 				route: [{observable: observableTypes.link, rel: rels.specialization}]
 			}
@@ -43,8 +45,8 @@ class ActivityDescriptionEditor extends HypermediaStateMixin(LitElement) {
 			<textarea class="d2l-input"
 				@input="${this._onInputDescription}"
 				placeholder="Write a description"
-				value="${this.description ? this.description : ''}"
-			></textarea>
+				.value="${this.description}"
+			>${this.description ? this.description : ''}</textarea>
 		</label>
 		` : null;
 	}
--- a/components/activity/description/d2l-activity-description-editor.js
+++ b/components/activity/description/d2l-activity-description-editor.js
@@ -26,12 +26,12 @@ class ActivityDescriptionEditor extends HypermediaStateMixin(LitElement) {
 		return [ inputLabelStyles, inputStyles,
 			css`
 			@media (max-width: 615px) {
-				textarea.d2l-input {
-					height: 100px;
+				.d2l-activity-description-editor textarea {
+					height: 5rem;
 				}
 			}
 
-			textarea.d2l-input {
+			.d2l-activity-description-editor textarea {
 				resize: none;
 			}
 			`
@@ -39,8 +39,8 @@ class ActivityDescriptionEditor extends HypermediaStateMixin(LitElement) {
 	}
 
 	render() {
-		return this.updateDescription.has ? html`
-		<label>
+		return this._hasAction('updateDescription') ? html`
+		<label class="d2l-activity-description-editor">
 			<span class="d2l-input-label">Description</span>
 			<textarea class="d2l-input"
 				@input="${this._onInputDescription}"
@@ -52,11 +52,10 @@ class ActivityDescriptionEditor extends HypermediaStateMixin(LitElement) {
 	}
 
 	_onInputDescription(e) {
-		if (this.updateDescription.has) {
+		if (this._hasAction('updateDescription')) {
 			this.updateDescription.commit({description: { observable: observableTypes.property, value: e.target.value} });
 		}
 	}
-
 }
 
 customHypermediaElement('d2l-activity-description-editor', ActivityDescriptionEditor);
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Review by zinabat — COMMENTED** [review]
This looks great overall!

File: components/activity/description/d2l-activity-description-editor.js [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+			}
+		};
+	}
+
+	static get styles() {
+		return [ inputLabelStyles, inputStyles,
+			css`
+			@media (max-width: 615px) {
+				textarea.d2l-input {
+					height: 100px;
```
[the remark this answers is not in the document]
m6jones: Yeah that be 4rem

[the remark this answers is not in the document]
dbatiste: Root font-size is `20px`... you sure you don't mean `5rem`?  😉

components/activity/description/d2l-activity-description-editor.js
```diff
@@ -0,0 +1,62 @@
...
+		];
+	}
+
+	render() {
+		return this.updateDescription.has ? html`
+		<label>
+			<span class="d2l-input-label">Description</span>
+			<textarea class="d2l-input"
+				@input="${this._onInputDescription}"
+				placeholder="Write a description"
```
AKobets, line 47: when we will add localization ?

lloydwaddell: Other components are missing localization as well, we should probably create a story for it. I didn't see a story in the backlog.

components/activity/description/d2l-activity-description-editor.js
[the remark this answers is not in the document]
lloydwaddell: would 4 or 5 rem be better? Judging by the UI design, I thought the equiv. row size was a 4. Didn't even know rem was a thing, so that is definitely handy.

components/activity/description/d2l-activity-description-editor.js
```diff
@@ -0,0 +1,62 @@
...
+			css`
+			@media (max-width: 615px) {
+				textarea.d2l-input {
+					height: 100px;
+				}
+			}
+
+			textarea.d2l-input {
+				resize: none;
+			}
```
lloydwaddell, line 36: Wondering, if I should change this to an ID as to not interfere with other textareas that could appear on the page.

lloydwaddell: The design doesn't show allowing resize, but if you add multiline text from mobile screen and then look at it on full screen you have to scroll. Thinking we should at least allow resizing for large screen sizes. Any thoughts?

components/activity/description/d2l-activity-description-editor.js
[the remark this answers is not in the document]
m6jones: 5 rem. I didn't do the math right :O

components/activity/description/d2l-activity-description-editor.js
m6jones: This is a question for Abbie

components/activity/description/d2l-activity-description-editor.js
m6jones: Yeah, we need to add localization to this repo Eek!

m6jones: (this is not for this PR)

m6jones (approved):

test/activity/description/d2l-activity-description-editor.test.js
```diff
@@ -0,0 +1,144 @@
...
+			const element = await _createComponentAndWait('/learning-path/existing');
+
+			let textarea = element.shadowRoot.querySelector('label textarea');
+			textarea.value = updatedDescriptionText;
+			const inputEvent = new Event('input');
+			textarea.dispatchEvent(inputEvent);
+
+			const spy = sinon.spy(element);
+			element._state.push();
+			await aTimeout(100);
```
AKobets, line 107: there is already timeout in _elementUpdated

test/activity/description/d2l-activity-description-editor.test.js
```diff
@@ -0,0 +1,144 @@
...
+async function _createComponentAndWait(path)
+{
+	const element = await fixture(html`<d2l-activity-description-editor href="${path}" token="test-token"></d2l-activity-description-editor>`);
+	await _elementUpdated(element);
+	return element;
+}
+
+// using a delay to wait for fetchMock to be called
+// would prefer a better way
+async function _elementUpdated(element) {
```
AKobets, line 20: do not hardcode `100` as a timeout and add millisecond parameter instead

lloydwaddell: added the delay parameter, and removed the extra aTimeout calls

test/activity/description/d2l-activity-description-editor.test.js
```diff
@@ -0,0 +1,142 @@
...
+			// update textarea
+			const textarea = element.shadowRoot.querySelector('label textarea');
+			textarea.value = updatedDescriptionText;
+
+			// create spy and trigger input event
+			const spy = sinon.spy(element.updateDescription);
+			const inputEvent = new Event('input');
+			textarea.dispatchEvent(inputEvent);
+
+			assert.isTrue(spy.commit.called, 'Commit should be called when input event is triggered');
```
AKobets, line 91: looks like some assertions are missing.
- What should change when commit is called?
- How to assert that save is not called ?

test/activity/description/d2l-activity-description-editor.test.js
```diff
@@ -0,0 +1,142 @@
...
+			const element = await _createComponentAndWait('/learning-path/existing');
+
+			const textarea = element.shadowRoot.querySelector('label textarea');
+			textarea.value = updatedDescriptionText;
+			const inputEvent = new Event('input');
+			textarea.dispatchEvent(inputEvent);
+
+			assert.equal(element.description, updatedDescriptionText, 'description should be updated');
+
+			await aTimeout(1000);
```
AKobets, line 128: is aTimeout needed here ?

lloydwaddell: yes, I've tried with lower timeouts even and the test would fail.

test/activity/description/d2l-activity-description-editor.test.js
lloydwaddell: Nothing changes in the component when commit is called, we only need to make sure that the commit function was fired.
We could spy on _state.push() and to verify it was not called.

AKobets: commit should update observers. what is observing the action?

AKobets: is it element.description that chould get the proper value?

test/activity/description/d2l-activity-description-editor.test.js
AKobets: it is not obvious what it is waiting for after dispatch event.  `element.description` is already asserted above. Should it wait for element updated ?

lloydwaddell: I fixed the issue by adding an additional await elementUpdated after firing the input event, the aTimeout isn't necessary anymore

test/activity/description/d2l-activity-description-editor.test.js
lloydwaddell: the element.description is where the property is stored, but it also reflects the value changed when the input event is fired.

lloydwaddell: I renamed the test, the "make sure it doesn't save" portion is covered in the reset test.

AKobets (approved):

PR #48, opened 2020-12-04
feat: updating collection editor to use sub entity objects

zinabat:
## Context
[US122956](https://rally1.rallydev.com/#/detail/userstory/460074259704?fdp=true)
Hooks up the `d2l-activity-editor-main-collection` component to support the new `subEntities` change. Other `subEntities` and `subEntity` types use routing for existing components, and the change does not affect them.

### DIFF
--- a/creating-new-components.md
+++ b/creating-new-components.md
@@ -103,7 +103,7 @@ render() {
 Types are based on Siren's hypermedia format.
 
 - `classes`: classes on the entity
-- `entity`: a dump of the entire entity object associated with this href
+- `entity`: the entire parsed entity object associated with this href
 - `link`: a string representing a link
 - `property`: a simple property that's part of the entity
 - `subEntity`: a parsed siren sub entity object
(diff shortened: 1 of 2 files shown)


creating-new-components.md
```diff
@@ -103,10 +103,11 @@ render() {
 Types are based on Siren's hypermedia format.
 
 - `classes`: classes on the entity
-- `entity`: an entity that's relevant to this entity, such as a course image
+- `entity`: a dump of the entire entity object associated with this href
```
AKobets, line 106: `a dump` what it means here?

m6jones (approved):

creating-new-components.md
zinabat: like a `var_dump` type thing - I can adjust the language

zinabat, line 106:
```suggestion
- `entity`: the entire parsed entity object associated with this href
```

PR #59, opened 2021-01-11
feat: Remove activity from a collection

zinabat:
## Context
[US123822](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F480212652368&fdp=true?fdp=true)

Allows removal of activities from a learning path.

Uses an event to do this to avoid an issue with `deepCopy`

## Quality
- Manual testing with `polarisdev`
- Tested adding an activity and then removing it
- Tested removing an existing activity
- Will not save until my save stuff is done/merged

### Known Issues
- There is an error coming from the `list-item-accumulator-mixin` that we need to fix elsewhere.


tenbradley:
:O It's happening!

components/activity/editor/collection/d2l-activity-editor-collection-add.js
```diff
@@ -226,12 +230,11 @@ class ActivityEditorCollectionAdd extends HypermediaStateMixin(LocalizeCollectio
 	}
 
 	_onAddActivityCommit() {
-		this._selectedCandidates.forEach(x => x.alreadyAdded = true);
```
zinabat, line 229: We don't need this because it updates when the items change.

components/activity/editor/collection/d2l-activity-editor-main-collection.js
```diff
@@ -93,14 +93,14 @@ class ActivityEditorMainCollection extends LocalizeFoundationEditor(SkeletonMixi
...
+						<d2l-activity-editor-collection-add href="${this._collectionHref}" .token="${this.token}">
 						</d2l-activity-editor-collection-add>
 
-						<div class="d2l-body-compact d2l-skeletize">${this.localize('text.activities')} ${this.items.length}</div>
+						<div class="d2l-body-compact d2l-skeletize">${this.localize('text.activities')} ${this._items.length}</div>
 					</div>
 				</div>
 				<div class="d2l-activity-collection-activities">
-					<d2l-list @d2l-list-item-position-change="${this._moveItems}">${repeat(this.items, item => item.href || item.properties.actionState, item => html`
+					<d2l-list @d2l-list-item-position-change="${this._moveItems}" @d2l-remove-collection-activity-item="${this._onRemoveActivity}">${repeat(this._items, item => item.href || item.properties.actionState, item => html`
```
zinabat, line 103: We're using an event to avoid attaching deep cloned items on each list item

AKobets (approved):
LGM

PR #60, opened 2021-01-11
feat: dialog box when saving an existing learning path fails

tenbradley:
Context:
[US123433](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F466974478716)
When saving a learning path fails, we give the user a dialog box asking them to attempt to save again.

Quality:
Testing using polarisdev sitelord.
- tested by throwing explicitly from the try statement
- need to test push actually throwing

### DIFF
--- a/components/activity/editor/d2l-activity-editor-footer.js
+++ b/components/activity/editor/d2l-activity-editor-footer.js
@@ -55,7 +55,7 @@ class ActivityEditorFooter extends LocalizeFoundationEditor(HypermediaStateMixin
 			<d2l-backdrop id="save-backdrop" for-target="#save-buttons" no-animate-hide ?shown="${this._backdropOpen}"></d2l-backdrop>
 			<d2l-dialog id="save-failed-dialog" ?opened="${this._dialogOpen}" @d2l-dialog-close="${this._closeDialog}" title-text="${this.localize('text-dialogSaveTitle')}">
 				<div>${this.localize('text-dialogSaveContent')}</div>
-				<d2l-button slot="footer" primary data-dialog-action="okay">${this.localize('label.ok')}</d2l-button>
+				<d2l-button slot="footer" primary data-dialog-action="okay">${this.localize('label-ok')}</d2l-button>
 			</d2l-dialog>
 			`;
 	}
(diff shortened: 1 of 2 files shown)


components/activity/editor/d2l-activity-editor-footer.js
```diff
@@ -42,22 +46,37 @@ class ActivityEditorFooter extends LocalizeFoundationEditor(HypermediaStateMixin
...
+			<d2l-alert-toast id="save-succeeded-toast" ?open="${this._toastOpen}" type="success"
+				announce-text="${this.localize('text.saveComplete')}">
+					${this.localize('text.saveComplete')}
 			</d2l-alert-toast>
-			<div><slot name="save-status">${this.localize('text.saveStatus')}</slot></div>
-			<d2l-backdrop id="save-backdrop" for-target="save-buttons" no-animate-hide></d2l-backdrop>
-		`;
+
+			<d2l-backdrop id="save-backdrop" for-target="#save-buttons" no-animate-hide ?shown="${this._backdropOpen}"></d2l-backdrop>
+			<d2l-dialog id="save-failed-dialog" ?opened="${this._dialogOpen}" @d2l-dialog-close="${this._closeDialog}" title-text="${this.localize('text.dialogSaveTitle')}">
```
zinabat, line 56: Are we thinking there will be multiple dialog components, or will we just change the language in them for the different save state failures/confirmations?

tenbradley: My hope would be to use the same dialog box and just change the language based on if it's a new learning path.
We concluded that we would determine if a learning path is new based on if it starts empty with default name and description, right? Is we could encapsulate that in a field once it's loaded, that would be awesome.

zinabat: Yeah that makes sense - should we rename the `id` of the dialog in that case?

tenbradley: Can do. I want something more specific than "backdrop", but less specific that "save-backdrop".

zinabat: I meant more for the `save-failed-dialog` like `information-dialog` or something

m6jones: The ids should be unique. https://github.com/BrightspaceUI/core/blob/master/helpers/uniqueId.js Don't worry about going back and changing all of them. I might do that later. Sorry just noticed this now.

zinabat (approved):
There's an issue unrelated to this pull request where the state will revert back to the original state before changes, but only after a failure happens and save is pressed again. Strange, not for this PR to fix (engine issue).

PR #67, opened 2021-01-13
Add localization to activity types

tenbradley:
Context:
Localization for labels was missed.
[US124010](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F483548927728)

### DIFF
--- a/components/activity/type/d2l-activity-type.js
+++ b/components/activity/type/d2l-activity-type.js
@@ -17,9 +17,9 @@ class ActivityType extends SkeletonMixin(LocalizeFoundationType(HypermediaStateM
 
 	static get components() {
 		return {
-			'learning-path': html`Learning Path`,
-			'course-offering': html`Course`,
-			default: html`Activity`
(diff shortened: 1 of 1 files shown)


components/activity/type/d2l-activity-type.js
```diff
@@ -17,9 +17,9 @@ class ActivityType extends SkeletonMixin(LocalizeFoundationType(HypermediaStateM
 
 	static get components() {
 		return {
-			'learning-path': html`Learning Path`,
-			'course-offering': html`Course`,
-			default: html`Activity`
+			'learning-path': 'label-learningPath',
```
AKobets, line 20: how youve tested this change?

tenbradley: I tried it out using the polarisdev the activities were receiving the Course label. That's the only type of activity we currently support.

m6jones (approved):

PR #68, opened 2021-01-13
Remove: UX tweaks

m6jones:
image

This this PR doesn't do:
Fix description styles.
Fix the background color which is being done on the LMS


zinabat (approved):
IT LOOKS SO GOOD!!!!

PR #69, opened 2021-01-14
feat: defactor localization files

tenbradley:
Context:
The localization.js files were nicely factored but this makes them not work with BSI.


components/activity/description/lang/localization.js
```diff
@@ -1,9 +1,62 @@
-import { getLocalizeResources } from '../../../../lang/localizeResources.js';
 import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
 
 export const LocalizeFoundationDescription = superclass => class extends LocalizeMixin(superclass) {
 
 	static async getLocalizeResources(langs) {
-		return getLocalizeResources(langs, 'components/activity/description');
+		let translations;
```
AKobets, line 6: currently if I try loading french it will throw exception  because the import of nonexisting file has failed.
I would suggest to have  try/catch statement and  return english translation  in case the import failed for unsupported locale.

tenbradley: We don't have the other language files yet so I do expect that to happen. A try catch could be used.

AKobets: this does not work unfortunately

AKobets (approved):

PR #70, opened 2021-01-14
feat: dialog box for save failing on a new learning path

tenbradley:
Context:
[US123434](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F466974498476)
We want text that is different for saving failing on a new learning path.

Quality:
Tested on polarisdev2.
- new text shows up when learning path is new
- editing text shows up when learning path is old


lloydwaddell (approved):
looks good to me

PR #71, opened 2021-01-14
infinite loop observing action in footer

lloydwaddell:
observing delete property is causing an infinite loop


m6jones:
Fixed but forgot to release when I did the merge. :O

PR #72, opened 2021-01-14
feat: Dialog component for load failures

zinabat:
## Context
[US123433](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F466974478716&fdp=true&fdp=true&fdp=true?fdp=true)

First portion of story. 
- Creates a new component for displaying a dialog when an activity load fails.
- Added to the `d2l-hc-activity-editor`
- Component is generic and will intelligently display langs based on the classes
- Supports generic `activity-usage` and `learning-path`
- Cancel just closes the dialog currently - will add redirecting in a different PR

Screen Shot 2021-01-14 at 1 28 37 PM

## Quality
- Manual testing by forcing the engine to throw an error
- There's a problem with the engine that's not catching errors correctly but is separate from this story

### DIFF
--- a/components/activity/dialog/d2l-activity-dialog-load-failed.js
+++ b/components/activity/dialog/d2l-activity-dialog-load-failed.js
@@ -1,7 +1,7 @@
 import '@brightspace-ui/core/components/dialog/dialog.js';
-import { css, LitElement } from 'lit-element/lit-element.js';
 import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
 import { html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
+import { LitElement } from 'lit-element/lit-element.js';
 import { LocalizeDialog } from './lang/localize-dialog.js';
 
 const getType = (classes) => {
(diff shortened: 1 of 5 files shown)


components/activity/dialog/d2l-activity-dialog-load-failed.js
```diff
@@ -0,0 +1,49 @@
+import '@brightspace-ui/core/components/dialog/dialog.js';
+import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
+import { LitElement } from 'lit-element/lit-element.js';
+import { LocalizeDialog } from './lang/localize-dialog.js';
+
+const getType = (classes) => {
+	if (classes.includes('learning-path')) return 'learning-path';
+	return 'activity-usage';
```
AKobets, line 9: should it check if classes has expected class for activity-usage ?

components/activity/dialog/d2l-activity-dialog-load-failed.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+class ActivityDialogLoadFailed extends HypermediaStateMixin(LocalizeDialog(LitElement)) {
+
+	static get properties() {
+		return {
+			_type: { observable: observableTypes.classes, method: getType },
```
m6jones, line 16: There not `type: String` (it might be `type: string`)

m6jones (approved):

components/activity/dialog/d2l-activity-dialog-load-failed.js
zinabat: I think leaving it as default is fine since otherwise langterms will break.

PR #77, opened 2021-01-14
fix: Style fixes for activity

m6jones:
image

Added languaged for activity count


zinabat (approved):
Tested on `polarisdev2`

PR #80, opened 2021-01-15
fix: Remove extra hashtag

zinabat:
## Context
Fixes Bug 6 caught by @AKobets 
**Before:** 6) Console logs exception when clicking Save And Close button
```
Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'DocumentFragment': '##save-buttons' is not a valid selector.
    at HTMLElement.updated (https://s.brightspace.com/lib/bsi/20.21.2-102/unbundled/backdrop.js:1:2147)
    at HTMLElement.performUpdate (https://s.brightspace.com/lib/bsi/20.21.2-102/unbundled/lit-element.js:1:16932)
    at HTMLElement._enqueueUpdate (https://s.brightspace.com/lib/bsi/20.21.2-102/unbundled/lit-element.js:1:16421)
```
(Error will be printed and then disappears. Enable 'preserve log upon navigation' in Dev tools Settings)

**After:** No bug because we're not sending a bad `id`

## Quality
Tested on `polarisdev2` with local BSI

### DIFF
--- a/components/activity/editor/d2l-activity-editor-footer.js
+++ b/components/activity/editor/d2l-activity-editor-footer.js
@@ -78,7 +78,7 @@ class ActivityEditorFooter extends LocalizeFoundationEditor(HypermediaStateMixin
 					${this.localize('text-saveComplete')}
 			</d2l-alert-toast>
 
-			<d2l-backdrop id="save-backdrop" for-target="#save-buttons" no-animate-hide ?shown="${this._backdropOpen}"></d2l-backdrop>
+			<d2l-backdrop id="save-backdrop" for-target="save-buttons" no-animate-hide ?shown="${this._backdropOpen}"></d2l-backdrop>
 			<d2l-dialog id="save-failed-dialog" ?opened="${this._dialogOpen}" @d2l-dialog-close="${this._closeDialog}" title-text="${this._isNew ? this.localize('text-newDialogSaveTitle') : this.localize('text-editDialogSaveTitle')}">
 				<div>${this._isNew ? this.localize('text-newDialogSaveContent') : this.localize('text-editDialogSaveContent')}</div>
 				<d2l-button slot="footer" primary data-dialog-action="okay">${this.localize('label-ok')}</d2l-button>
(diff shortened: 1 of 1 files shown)


components/activity/editor/d2l-activity-editor-footer.js
```diff
@@ -78,7 +78,7 @@ class ActivityEditorFooter extends LocalizeFoundationEditor(HypermediaStateMixin
 					${this.localize('text-saveComplete')}
 			</d2l-alert-toast>
 
-			<d2l-backdrop id="save-backdrop" for-target="#save-buttons" no-animate-hide ?shown="${this._backdropOpen}"></d2l-backdrop>
+			<d2l-backdrop id="save-backdrop" for-target="save-buttons" no-animate-hide ?shown="${this._backdropOpen}"></d2l-backdrop>
```
m6jones, line 81: could you make this use unique ids?

m6jones: id="${this.saveButtons}

zinabat: I forgot about this - done.

I removed all of the extraneous ids because I'm 95% sure they are not needed. I also fixed a `z-index` issue.

components/activity/editor/d2l-activity-editor-footer.js
```diff
@@ -49,6 +49,7 @@ class ActivityEditorFooter extends LocalizeFoundationEditor(HypermediaStateMixin
 			}
 			.d2l-activity-editor-save-buttons {
 				display: flex;
+				z-index: 1000;
```
zinabat, line 52: Allows the buttons to show up from behind the backdrop.

dbatiste: What's the background for needing a special backdrop (since dialog already provides one)?  

I'm a bit worried that this sorta thing will need to be repeated since and most times when `z-index` is used, it takes some sleuthing to figure out why a certain value was chosen, and it also makes changing `z-index` values super risky.  In this case I _think_ it's because the backdrop uses 999, dialogs use 1000, and there are a variety of things that are either required to be in front of, or behind.  If we really need to do this, I think this z-index should either be applied automatically by the backdrop, or we should provide some sharable CSS.

tenbradley: The backdrop is supposed to be visible while we are waiting for the save to complete to prevent the user from modifying the page.

AKobets (approved):
tested.

PR #89, opened 2021-01-21
feat: display question type text

stevenao:


mdgbayly (approved):
Ok with this going in. Might want to change if we decide to represent the type information as a sub-entity in the API.

PR #92, opened 2021-01-26
fix: footer appears behind browse library dialog

tenbradley:
### Context:
[DE42228](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F486927762512)
In firefox, the save and close, cancel, and visibility are appearing in front of the shim from the collection dialog.

### Quality:
Tested on polarislocalbsi in chrome and firefox.

### DIFF
--- a/components/activity/editor/d2l-hc-activity-editor.js
+++ b/components/activity/editor/d2l-hc-activity-editor.js
@@ -60,7 +60,6 @@ class ActivityEditor extends LitElement {
 				left: 0;
 				position: fixed;
 				right: 0;
-				z-index: 999;
 			}
 			.d2l-activity-editor-template-default-footer * {
 				margin: auto;
--- a/components/activity/editor/d2l-activity-editor-footer.js
+++ b/components/activity/editor/d2l-activity-editor-footer.js
@@ -49,7 +49,6 @@ class ActivityEditorFooter extends LocalizeFoundationEditor(HypermediaStateMixin
 			}
 			.d2l-activity-editor-save-buttons {
 				display: flex;
-				z-index: 999;
 			}
 			.d2l-desktop-button {
 				display: none;


components/activity/editor/d2l-hc-activity-editor.js
```diff
@@ -60,7 +60,6 @@ class ActivityEditor extends LitElement {
 				left: 0;
 				position: fixed;
 				right: 0;
-				z-index: 999;
```
m6jones, line 63: Removing z-index all together, won't that cause those other issues with things popping above and the like?

AKobets: we need to test  it before merge. I will test it on windows chrome and firefox.  @maya, could you please check it on safari ?

AKobets: > we need to test it before merge. I will test it on windows chrome and firefox. @maya, could you please check it on safari ?

Verified that there is no regression on windows chrome, edge and firefox.

m6jones: Does local bsi work on edge now? or did you build it? If you built it could you upload the build to s3?

m6jones: As per removing the z-index. I see, the cool new scroll bars allows us to do it! Nice Ten!

I prefer to test safari after we put it in bsi. Setting up safari dev build will take half a day :( I don't see why this wouldn't work in safari. I'll eat my words and work into the night if I am wrong.

m6jones:
I approve but I want to make sure Alex approves my comment I made.

components/activity/editor/d2l-hc-activity-editor.js
tenbradley: In theory it could, with playing with it, I didn't see any issues. I know Alex had played around with z-index stuff last week.

tenbradley: Yay scroll bar!!!

tenbradley: In theory, setting the z-index to 998 should work since the z-index for backdrop is 999. But why do we have a z-index here anyways?

AKobets (approved):

components/activity/editor/d2l-hc-activity-editor.js
AKobets: > Does local bsi work on edge now? or did you build it? If you built it could you upload the build to s3?

I was not accurate. It worked for me in Chromium Edge with local bsi

PR #94, opened 2021-01-26
fix: Removed aria-live to prevent SR reading changes

lloydwaddell:
[DE42282](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F489328036512)

```Context```
Clicking the load more button would trigger the screen reader to read each new activity name as they were loaded. Skeleton items were also being read as "name". This caused some spam.

The dialog was created as an aria-live region that will announce changes, so I've removed the attribute. 

```Quality```
Tested using NVDA screen reader
1) Clicking load more no longer causes activities to be announced
2) Tabbing through the new activities will continue to announce the name as they receive focus


m6jones:
Could you update the skeleton (default component) and replace Name with, Item is loading?

m6jones:
It needs text in there to get the right dimensions and That seems a little more useful if other people hit it.

m6jones (approved):

PR #100, opened 2021-02-05
feat: create test utility functions

tenbradley:
### Context:
We need to make it easy for developers to test their components. This pr is refactors some functions from the description test to a utility file to be used for other components moving forward.
It also cleans up the test directory a bit my making a data directory.

### Quality:
Updated description tests to use util functions and tests pass.


test/test-util.js
```diff
@@ -0,0 +1,30 @@
...
+// using a delay to wait for fetchMock to be called
+// would prefer a better way
+export async function delayAndAwaitForElement(element, delayMs) {
+	return aTimeout(delayMs).then(async() => {
+		await elementUpdated(element);
+	});
+}
+
+// Handle updating the component textarea with a new value and await the update
+export async function fireTextareaInputEvent(element, label, updatedDescriptionText) {
```
AKobets, line 21: textarea, label, updatedDescriptionText should be renamed to be more generic.
Also may be 'input' should be parametrized. Could event have data ?

tenbradley: I agree I should update it to be more generic. I started with just a copy and paste for now as I am not sure in which ways we'll want it to be more generic.

PR #105, opened 2021-02-11
test: function to make adding mock data for tests easy

tenbradley:


test/data/fetchMocks.js
```diff
@@ -18,31 +18,16 @@ function GenerateComponentLink(linkPath) {
...
-	.mock('path:/course/existing', () => {
-		return GenerateComponentLink('/course/existing/object');
-	})
-	.mock('path:/course/existing/object', () => {
-		return courseExisting;
-	});
+export function addToMock(path, object) {
+	mockLink.mock(`path:${path}`, () => {return GenerateComponentLink(`${path}/object`);});
+	if (object) {
+		mockLink.mock(`path:${path}/object`, () => {return object;});
```
zinabat, line 31:
```suggestion
		mockLink.mock(`path:${path}/object`, () => object);
```

test/testTemplate.js
```diff
@@ -0,0 +1,76 @@
...
+		it('should ...', async() => {
+			const element = await _createComponent();
+
+			// paths should be followed
+			assert.isTrue(mockLink.called('path:/learning-path/new'), '/learing-path/new was not called');
+			assert.isTrue(mockLink.called('path:/learning-path/new/object'), '/learing-path/new/object was not called');
+
+			// call methods on element and ensure expected results
+			assert(element);
+		});
```
lloydwaddell, line 73: Could this be merged into a common method called SetupMock(path, object) that creates the mock and validates this test?

tenbradley: Hmm, probably. The issue is that it might not call fetch on both objects, so I am hesitant to hide those calls. I also think it is easier to read the test is all asserts are in the test.

lloydwaddell: It should always call both objects though, this is basically just testing that the mock was setup properly and an element was returned. If it fails its either a problem with our testing framework or the engine, unless I am missing something else?

tenbradley: Maybe? There was a test I was writing where only one was called, but that might have been before I had finished the test and it might have called both in the end....

test/data/fetchMock.js
```diff
@@ -0,0 +1,39 @@
...
+				href: linkPath,
+				rel: ['https://api.brightspace.com/rels/organization', 'https://api.brightspace.com/rels/specialization']
+			}
+		]
+	};
+}
+
+export const mockLink = fetchMock;
+
+export async function addToMock(path, object, createComponent, testObject = true) {
```
tenbradley, line 23: This needs to be cleaned up and made more generic.

PR #107, opened 2021-02-18
feat: activity editor defaults to primary-secondary template

tenbradley:
### Context: 
[US122674](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F458814721568)
The learning path editor needs to be switched to use primary-secondary to accommodate additional information. This pr is updating the editor to use the template but does not use the secondary area yet.
Some CSS formatting still needs to be done.

primary-secondary

### Quality:
- tested locally in sitelord with local bsi

### DIFF
--- a/components/activity/editor/d2l-hc-activity-editor.js
+++ b/components/activity/editor/d2l-hc-activity-editor.js
@@ -24,23 +24,10 @@ class ActivityEditor extends LitElement {
 			:host {
 				display: block;
 			}
-			.d2l-activity-editor-template-primary-secondary {
-				display: grid;
-				grid-template-areas:
-					"header"
-					"content"
-					"footer";
-				grid-template-columns: auto;
-				grid-template-rows: auto 1fr auto;
-				height: calc(100vh - 106px);
-				margin: auto;
-				max-width: 1230px;
-				overflow-y: scroll;
+			.d2l-primary-secondary {
 				position: relative;
-				border: none;
 			}
-
-			.d2l-activity-editor-template-default {
+			.d2l-activity-editor-template {
 				display: grid;
 				grid-template-areas:
 					"header"
@@ -112,7 +99,7 @@ class ActivityEditor extends LitElement {
 
 	render() {
 		const templates = {
-			'default': () => this._renderDefault(),
+			'single-pane': () => this._renderSinglePane(),
 			'primary-secondary': () => this._renderPrimarySecondary()
 		};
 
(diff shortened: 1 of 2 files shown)


components/activity/editor/d2l-hc-activity-editor.js
```diff
@@ -93,7 +106,7 @@ class ActivityEditor extends LitElement {
 
 	constructor() {
 		super();
-		this.template = 'default';
+		this.template = 'primary-secondary';
```
lloydwaddell, line 109: not sure we want to commit, maybe just leave it as a local change until the LMS is updated?

components/activity/editor/d2l-hc-activity-editor.js
```diff
@@ -24,6 +24,20 @@ class ActivityEditor extends LitElement {
 			:host {
 				display: block;
 			}
+			.d2l-activity-editor-template-primary-secondary {
```
lloydwaddell, line 27: this shares common css with the default template, we could create a general class and then only override what is necessary on the specific templates

lloydwaddell:
There is a light grey border or background bleeding through on the right side, do you what is causing this issue?
border-color

components/activity/editor/d2l-hc-activity-editor.js
tenbradley: I'm also not sure. I think more activities editors are going to use the primary-secondary template, so that being the default might be a good idea. Just seems weird because the other one is named default. Renaming it to primary, or something similar, might be a good idea.

lloydwaddell: I wonder if there is a way to play with css and slots to only show the secondary panel if something has been added to it.

tenbradley:
> There is a light grey border or background bleeding through on the right side, do you what is causing this issue?
> border-color

I'm pretty sure that's just part of the primary-secondary template?

components/activity/editor/d2l-hc-activity-editor.js
tenbradley: I agree. I just wasn't sure how to best to this.

components/activity/editor/d2l-hc-activity-editor.js
tenbradley: Maybe? But also, that assumes that is what design wants.

components/activity/editor/d2l-hc-activity-editor.js
lloydwaddell: I would create another class d2l-activity-editor-template and it to both outer wrappers with the shared css, and have the -default and -primary-secondary classes have the overridden values

lloydwaddell (approved):

PR #114, opened 2021-03-01
test: reorganize tests to be in component directory

tenbradley:
### Context:
We want tests to be in component directories. This pr is moving the tests we currently have.


zinabat (approved):
ugh this is so much better.

PR #115, opened 2021-03-01
feat: Convert discover rule picker components to foundations 

zinabat:
## Context
[US123170](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F462626337568&fdp=true?fdp=true)

- Converted the `d2l-discover-rule-picker-dialog` and `d2l-discover-rule-picker` components to be foundation components from these: https://github.com/Brightspace/Discover-Components/tree/master/components
- Accessibility improvements
- Adds tests for functionality
- Commits an action when done is pressed

### What's left?
- This is still a pretend API, so some responses may need adjustment
- Conditions text input will become the new attribute-picker by @BhamelinD2L
- Conditions will be adjusted to take multiple values

## Quality
- New tests based on the old ones: https://github.com/Brightspace/Discover-Components/tree/master/test
- No tests added for `d2l-discover-rules` because that component doesn't do anything that would require testing


features/discover/test/d2l-discover-rule-picker-dialog.test.js
```diff
@@ -0,0 +1,120 @@
...
+	],
+	links: [
+		{ rel: ['self'], href: selfHref }
+	]
+};
+
+describe('d2l-discover-rule-picker-dialog', () => {
+
+	before(() => {
+		fetchMock.mock(selfHref, JSON.stringify(entity));
```
zinabat, line 39: I'd love to be able to properly reset these, but can't because other tests seem to rely on pre-created mocks that would be destroyed.

features/discover/d2l-discover-rule-picker-dialog.js
```diff
@@ -0,0 +1,103 @@
...
+		const picker = this.shadowRoot.querySelector('d2l-discover-rule-picker');
+		this._state.updateProperties({
+			conditions: {
+				observable: observableTypes.subEntities,
+				rel: rels.condition,
+				value: picker.conditions
+			}
+		});
+		// action is commited differently because it's a JSON string
+		this.updateConditions.commit({
```
zinabat, line 95: I have no idea what this action will actually look like so this is just straight up JSON for now.

features/discover/d2l-discover-rules.js
```diff
@@ -11,7 +11,7 @@ const rels = Object.freeze({
 	rule: 'rule',
 	entitlementRules: 'entitlement-rules',
 	conditionType: 'condition-type',
-	conditionTypes: 'available-condition-types'
+	newRule: 'new-rule'
```
BhamelinD2L, line 14: I'm a little confused about rule, entitlementRules, and newRule. 
is rule vs newRule the before/after states?  what if there are multiple rules being changed?

BhamelinD2L (approved):
Looks good!

features/discover/d2l-discover-rules.js
zinabat: This is where I floundered - How do we know what a new rule should look like before it's been created? This has been a consistent problem with activity creation workflows, so I decided to try something new.

`new-rule` is a link that gives the client a "template" of sorts on what a new-rule looks like and the actions it can perform. As it stands this is pretty minimal, because the workflow is new. I welcome feedback on alternatives to doing this, but we discuss it almost every time there's an Activity Model sync up.

`rule` is a rule that has already been created and exists somewhere. I'm... not positive how we're going to to this without actually creating it on the server. The correct solution is probably to use working copy, I'm unsure. I think I need @m6jones 's eyes on this bit

zinabat: Sticking this here for context:

I synced up with Maya and she agrees with the idea of the `new-rule` as a template that informs the UI of what new entities should look like/what they can do.

Other thoughts we had:
- it is a common workflow to create a new entity by opening a dialog or loading a different component and manipulating that entity before hitting a "done" button.
- we can add support in the engine for doing this **without working copy** because some features will not have working copy in scope or otherwise be unable to implement it.
- in the engine, we can create a workflow that attaches the subentity (in this case the `new-rule`) to the parent as a new subentity. We can give it a unique ID to be used in place of the normal `href`
- Doing this will allow us to use foundation components for subentities that haven't actually been created yet on the server

PR #119, opened 2021-03-03
fix: Update eslint configuration

lloydwaddell:
## Context
`lint` was crashing when we tried to switch to dynamic imports because we were using an old version of `babel-parser`, because our lint config was 6 versions behind.

So we fixed it so we can use danny's fancy thing now


test/.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-	"extends": "brightspace/open-wc-testing-config"
+	"extends": "brightspace/open-wc-testing-config",
+  "rules": {
```
lloydwaddell, line 3: nit: mixing white space

m6jones (approved):

PR #120, opened 2021-03-03
doc: move demo files to component directory

tenbradley:
### Context:
[US125516](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F505569581392&fdp=true?fdp=true)
We moved tests, now let's move the demos!

### Quality:
Checked demo pages after update.
- activity code is not displaying. Probably based on being in progress.
- new learning path data not showing up as expected.


components/activity/editor/demo/d2l-activity-editor-header.html
```diff
@@ -16,12 +16,12 @@
 		<d2l-demo-page page-title="Editor Header">
 			<h2>New Learning Path</h2>
 			<d2l-demo-snippet>
-				<d2l-activity-editor-header href="../data/base/new-learning-path.json" token="cake"></d2l-activity-editor-header>
+				<d2l-activity-editor-header href="/demo/data/base/new-learning-path.json" token="cake"></d2l-activity-editor-header>
```
lloydwaddell, line 19: json file paths need to be updated

demo/data/base/new-learning-path.json
```diff
@@ -29,18 +29,12 @@
...
       "href": "https://activities.api.proddev.d2l/activities/6606_706000_6607/usages/6607/users/169"
     },
-    {
-      "rel": [
-        "https://api.brightspace.com/rels/organization",
-        "https://api.brightspace.com/rels/specialization"
-      ],
-      "href": "../../../demo/data/base/new-learning-path.json"
-    },
+
```
lloydwaddell, line 32: We need this to make demos work for name/desc/code, all those properties are being referenced off of the organization. This file seems to be serving as both activity usage and organization, which is likely my fault during the initial setup. We may want to cleanup the data files to mimic the APIs responses better.

lloydwaddell: For now we could just put this line back in and target the data files in the next PR

m6jones: I would love for the API responses to be generic to fit the exact pattern of the component instead of matching our exact API. That way people can see how to use it and what they need to add to their API to make the component work.

tenbradley: Sounds like a reasonable thing. Would it make sense to have a separate story for that?

lloydwaddell: Yeah, I think a bigger story may be needed to make sure we are handling data and routing properly

m6jones (approved):

PR #123, opened 2021-03-04
fix: change divider colour for caketray to match design

tenbradley:
### Context: 
[US122674](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F458814721568&fdp=true?fdp=true)
The horizontal line is not the right colour currently. This updates it.

### DIFF
--- a/components/activity/editor/collection/d2l-activity-editor-secondary-card.js
+++ b/components/activity/editor/collection/d2l-activity-editor-secondary-card.js
@@ -41,7 +41,7 @@ class ActivityEditorSecondaryCard extends LitElement {
 				grid-row: content;
 			}
 			hr.d2l-split {
-				border-bottom: solid 1px redvar(--d2l-color-corundum)
+				border-bottom: solid 1px var(--d2l-color-corundum)
 			}
 		`];
 	}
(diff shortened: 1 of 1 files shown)


components/activity/editor/collection/d2l-activity-editor-secondary-card.js
```diff
@@ -53,7 +56,7 @@ class ActivityEditorSecondaryCard extends LitElement {
 		<d2l-labs-accordion-collapse flex>
 			<span slot="header">
 				<h3>${this.titleText}</h3>
-				<hr>
+				<hr class="split">
```
zinabat, line 59:
```suggestion
				<hr class="d2l-split">
```
Custom classes like this are always prefixed with `d2l-` by convention. Should probably change the other ones too.

lloydwaddell, line 59: this needs to happen on the default as well

tenbradley: It has been done.

tenbradley: Actually, it only needs to be in the default one. The accordion the other is using has a line built in which I have switched it to use now.

zinabat (approved):

tenbradley:
Closing as the card has been moved to a ui-labs repo and this is unneeded.

PR #124, opened 2021-03-04
fix: Add sample button for legacy page test

zinabat:
## Context
[US123769](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F480099321508&fdp=true&fdp=true?fdp=true)

Attempting to include a bare basic webcomponent in a legacy page via the "double adaptor" strategy @BhamelinD2L has talked about. See story for details.

## Quality
Added to demo page but tests unnecessary since this is a temporary component. `fix` prefix is to make sure we bump the version.

### DIFF
--- a/features/discover/d2l-discover-sample-button.js
+++ b/features/discover/d2l-discover-sample-button.js
@@ -1,0 +1,35 @@
+/**
+ * This button is designed for demoing with a legacy page and will be removed once the rules
+ * component is integrated into the Course Offering Information page
+ */
+import '@brightspace-ui/core/components/button/button.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { LocalizeDiscoverEntitlement } from './lang/localization.js';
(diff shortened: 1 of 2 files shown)


features/discover/d2l-discover-sample-button.js
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	render() {
+		return html`
+			<d2l-button @click="${this._onClick}">${this.text}</d2l-button>
+		`;
+	}
+
+	_onClick() {
+		const event = new CustomEvent('d2l-sample-button-click', {
```
zinabat, line 26: Realistically we won't actually need to use events in the real component, but I want to make sure this works.

BhamelinD2L: That's what I had in mind as well, thank you. Definitely worth doing to really explore how this works and hopefully get some practical documentation for it somewhere, other than the current buried slack messages 😄

BhamelinD2L (approved):
This does everything it should do, thanks!

PR #125, opened 2021-03-05
fix: moved code component tests

lloydwaddell:
https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F505772305740

```Context```
tests should live in the component root

```Quality```
unit tests passing

### CODE REVIEW

File: components/activity/description/test/d2l-activity-description-editor.test.js [hunk 1]
```diff
@@ -26,9 +26,9 @@ describe('d2l-activity-description-editor', () => {
 
 	before(async() => {
 		mockLink.reset();
-		await addToMock('/learning-path/new', learningPathNew, _createDescriptionEditor);
+		await addToMock('/learning-path/new', learningPathNew, _createDescriptionEditor, false);
```
lloydwaddell, line 29: the addMock assertion was failing had to turn it off to make tests pass. It complained that the object path was not called during the test creation. I think we might need to revisit how addMock works.

tenbradley (approved):

PR #126, opened 2021-03-05
fix: add aria-label to code input

lloydwaddell:
https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F459180378604

```Context```
Screen reader was not matching expectation in test docs

```Quality```
screen reader should read textfield {label} {description} {value}
tested using NVDA

### DIFF
--- a/components/activity/code/custom/d2l-activity-code-editor-learning-path.js
+++ b/components/activity/code/custom/d2l-activity-code-editor-learning-path.js
@@ -3,6 +3,7 @@ import '@brightspace-ui/core/components/inputs/input-text.js';
 import { css,  LitElement } from 'lit-element/lit-element.js';
 import { customHypermediaElement, html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
 import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
 import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-label-styles.js';
 import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/activity/code/custom/d2l-activity-code-editor-learning-path.js [hunk 1]
```diff
@@ -41,21 +42,23 @@ class ActivityCodeEditorLearningPath extends LocalizeDynamicMixin(HypermediaStat
...
 				<div class="d2l-activity-code-editor-description">${this.localize('text-code-description-lp')}</div>
-				<d2l-input-text
-					@input="${this._onInputCode}"
-					placeholder="${this.defaultLearningPathCode}"
-					value="${this.code}"
-					?skeleton="${!this._loaded}"
-					maxlength=50
-				></d2l-input-text>
 			</label>
+			<d2l-input-text
```
m6jones, line 54: d2l-input-text has a built in label?

dbatiste: Indeed - I think all of our core inputs provide built-in labels because the actual `input` element is inside the shadowDOM, and id-refs cannot cross dom-contexts / shadow-boundaries.  So currently, the `for` attribute above is not doing anything a11y-wise because it's not referencing the actual `input`.  With `d2l-input-text`, it is better to provide the label on `d2l-input-text` itself.  There is a `label-hidden` property that can be set if a hidden label is required.

There is ongoing work in the specs to allow x-dom id-refs to support this exact case, but nothing that we can rely on currently.

lloydwaddell: I see that, but my design requires inline description between the label and the input. Is this possible with the current d2l-input-text?

dbatiste: I see.  So currently the component doesn't support the label and the input inline.  

But yeah, you latest changes + `aria-hidden="true"` on the `div` so that the text doesn't get announced twice to screenreader.  Unfortunately, this means that when the user clicks on the visible text, focus won't be moved to the input... which is a standard behaviour for labels and inputs.

m6jones (approved):

PR #128, opened 2021-03-06
Demonstrating weird broken test

zinabat:
For some reason, a test that was working before breaks when localization is changed over to use the dynamic mixin.
Adding a check to the engine fixes this, but I couldn't reproduce it in the engine tests >:/


zinabat:
Good news is, this was a problem before dynamic langs: https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/130

PR #130, opened 2021-03-08
test: Adding features to sauce config

zinabat:
I expect at least one thing to break.


tenbradley:
> I expect at least one thing to break.
This is one of the most realistic messages I have seen.

PR #136, opened 2021-03-09
fix: Fixing rule picker cancel

zinabat:
## Context
[US123170](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F462626337568&fdp=true?fdp=true)

The rule picker dialog should remove changes from the dialog if cancel is pressed.
I could have sworn I had this working before. Anyway, it is fixed now.

### DIFF
--- a/features/discover/d2l-discover-rule-picker-dialog.js
+++ b/features/discover/d2l-discover-rule-picker-dialog.js
@@ -70,7 +70,6 @@ class RulePickerDialog extends LocalizeDiscoverEntitlement(HypermediaStateMixin(
 	}
 
 	_onCancelClick() {
-		this._copiedConditions;
 		this.requestUpdate().then(() => {
 			const picker = this.shadowRoot.querySelector('d2l-discover-rule-picker');
 			picker.reload(this._copiedConditions);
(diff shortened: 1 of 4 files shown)


features/discover/d2l-discover-rule-picker.js
```diff
@@ -88,8 +88,9 @@ class RulePicker extends LocalizeDiscoverEntitlement(HypermediaStateMixin(RtlMix
 		}
 	}
 
-	reload(newConditions) {
+	async reload(newConditions) {
 		this.conditions = newConditions;
```
BhamelinD2L, line 92: I found that changing the fields, then cancelling, seems to work correctly the first time. Doing it a second time does not cancel the change.
I think when the dialog is calling `_onCancelClick_` you are setting `this.conditions = this._copiedConditions`  and then feeding it to this reload function. This is letting your `_copiedConditions_ backup get mutated. I think we need to create a copy of the backup somewhere.

zinabat: You're right. I got around this by just not setting the conditions in the dialog. Please test again.

BhamelinD2L: Your change appears to work, until the Done button is pressed. After that, cancel longer cancels. I think it's grabbing the condition references back from the picker there.

zinabat: Oh right - this was actually intentional because we'll need to support editing in the future, but I'll get rid of it for now.

zinabat: copying from slack: I made it so that it will support editing new rules. I’ve changed it so that it will clear the dialog completely in cases where we’re creating a brand new rule. I even wrote a test to prove it works >:[

features/discover/d2l-discover-rule-picker-dialog.js
```diff
@@ -84,13 +83,24 @@ class RulePickerDialog extends LocalizeDiscoverEntitlement(HypermediaStateMixin(
...
 		const picker = this.shadowRoot.querySelector('d2l-discover-rule-picker');
-		this._state.updateProperties({
-			conditions: {
-				observable: observableTypes.subEntities,
-				rel: rels.condition,
-				value: picker.conditions
-			}
-		});
+		if (this.creating) {
+			const event = new CustomEvent('d2l-discover-rule-created', {
```
zinabat, line 87: we create an event for right now to tell the collection there's a new rule. Probably this should actually be attached to the new rule entity itself, I will see.

features/discover/test/d2l-discover-rule-picker-dialog.test.js
```diff
@@ -33,10 +34,23 @@ const entity = {
 	]
 };
 
+const newEntity = {
```
zinabat, line 37: Need a separate testing entity for new rules.

BhamelinD2L (approved):
This solution makes sense, and the test is appreciated. Cool!

PR #141, opened 2021-03-11
Making data files represent the APIs

lloydwaddell:
```Context```
Started to mock API endpoint data for use with demo and testing. The data we were using to demo the activity editor was recently broke by removing the organization link. In my last attempt to fix it I replaced the link pointing the file to itself, which ended up causing the engine to loop until it failed.

This is my first attempt at creating a more realistic representation of the data we would receive from the APIs.

```Quality```
Demo no longer throws an error, but does not display due to a CSS issue that breaks the demo page.

```Questions```
Considering moving data to its own folder and making it shareable with tests

### DIFF
--- a/demo/data/base/activity-usages/new-learning-path.json
+++ b/demo/data/base/activity-usages/new-learning-path.json
@@ -1,0 +1,31 @@
+{
+  "class":["activity-usage","learning-path","draft-published-entity","draft"],
+  "links":[
+    {
+      "rel":["https://activities.api.brightspace.com/rels/activity-usage","self"],
+      "href":"/demo/data/base/activity-usages/new-learning-path.json"
+    },
+    {
+      "rel":["https://activities.api.brightspace.com/rels/activity-collection"],
+      "href":"/demo/data/base/collection-empty.json"
+    },
+    {
+      "rel":["https://api.brightspace.com/rels/organization","https://api.brightspace.com/rels/specialization"],
+      "href":"/demo/data/base/organizations/new-learning-path.json"
+    }
+  ],
+  "actions": [
+    {
+      "href":"/activity-usage/update/draft",
+      "name":"update-draft",
+      "method":"PATCH",
+      "fields":[{"type":"checkbox","name":"draft","value":true}]
+    },
+    {
+      "href":"/activity-usage/associations",
+      "name":"query-associations",
+      "method":"GET",
+      "fields":[{"type":"text","name":"type"}]
+    }
+  ]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: demo/data/base/activity-usages/new-learning-path.json [hunk 1]
```diff
@@ -0,0 +1,31 @@
+{
+  "class":["activity-usage","learning-path","draft-published-entity","draft"],
```
AKobets, line 2: `"draft-published-entity","draft"` - could both classess be together? what states they represent?

demo/data/base/organizations/new-learning-path.json
```diff
@@ -0,0 +1,55 @@
...
+      "name":"update-draft","method":"PATCH",
+      "fields":[{"type":"checkbox","name":"draft","value":true}]
+    },
+    {
+      "href":"/organizations/update/code",
+      "name":"update-code","method":"PATCH",
+      "fields":[{"type":"text","name":"code","value":"LP"}]
+    },
+    {
+      "href":"/organizations/de;ete",
```
AKobets, line 50: `de;ete` - looks like a typo

tenbradley:
In response to your question, I think it makes sense to have one folder with all the data for tests and demos and had thought of that while making/moving them both.

demo/data/base/activity-usages/new-learning-path.json
lloydwaddell: draft-published-entity signifies that the draft field can be edited for this activity-usage (really organization)
draft is the current state and can be replaced by published when its been changed

zinabat:
> In response to your question, I think it makes sense to have one folder with all the data for tests and demos and had thought of that while making/moving them both.

I disagree with this statement, but it depends on what we end up deciding for test organization.

Tests are easier to read when the data is available within the testing file itself, which is why the engine integration tests and the tests in the discover feature folder are written that way. Data for demos functions as a full integration between the data response and the demo, where tests are testing very specific functionality based on usually a much smaller part of the data response.

lloydwaddell:
> I could be convinced that the duplication of response data is counter productive.

I think having specific test data in the feature folders themselves makes sense. My thinking for this was a fairly generic sample set of data that would be freebies for anyone to use. Then to Zina's point any specialized data needed should be made in the feature folders to not break other tests when a modification needs to be made.

Personally, I find having demos and tests in the folders of the components harder to work with during development. But I am willing to accept I am wrong lol.

PR #144, opened 2021-03-17
feat: open edit question fullscreen dialog

stevenao:
- to wire up open fullscreen dialog method on item click
- to refresh item and collection when return from edit

### DIFF
--- a/components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
+++ b/components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
@@ -7,6 +7,7 @@ import './d2l-activity-collection-item-quiz.js';
 import { css, LitElement } from 'lit-element/lit-element.js';
 import { customHypermediaElement, html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
 import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { ifDefined } from 'lit-html/directives/if-defined';
 import { repeat } from 'lit-html/directives/repeat';
 
 const rels = Object.freeze({
@@ -25,6 +26,11 @@ class ActivityEditorMainCollection extends HypermediaStateMixin(LitElement) {
 				route: [
 					{ observable: observableTypes.link, rel: rels.collection }
 				]
+			},
+			_organization: {
+				type: String,
+				observable: observableTypes.link,
+				rel: 'https://api.brightspace.com/rels/organization'
 			}
 		};
 	}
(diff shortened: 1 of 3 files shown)


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -71,14 +77,23 @@ class ActivityEditorMainCollection extends HypermediaStateMixin(LitElement) {
...
+						<d2l-activity-collection-item-quiz number="${idx + 1}" href="${item.href}" .token="${this.token}" key="${item.properties.id}" ou="${ifDefined(this._getOrgUnitId() || undefined)}"></d2l-activity-collection-item-quiz>
 						`)}
 					</d2l-list>
 				</div>
 			</div>
 		`;
 	}
 
+	_getOrgUnitId() {
+		if (this._organization) {
```
mdgbayly, line 89: This feels very hacky. I wonder if there are alternate approaches we could take. 
One option would be to just accept that we have to get the orgunitid from somewhere but try and isolate that as much as possible. I think it is possible to get it from the document element now. I think the HTML editor integration is using some trick like that. 

But possibly a better approach, which is more work, would be to use something like the ideas we were discussing yesterday for creating opaque slugs that capture all the information we need. So for example we could make our collection item identifier capture both the object id and the org unit id if we need both.

stevenao: yeah, I wasn't sure about it too. The idea was originated from `d2l-activities` - https://github.com/BrightspaceHypermediaComponents/activities/blob/18082b754c527156511edb362e5916bba43dfcb9/components/d2l-activity-card/d2l-activity-card.js#L291. 

I will look into your suggestions.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -9,6 +9,8 @@ import { html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedi
 
 import { ListItemLinkMixin } from '@brightspace-ui/core/components/list/list-item-link-mixin.js';
 
+const context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
```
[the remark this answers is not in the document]
mdgbayly: Yeh - i agree - not perfect either to use the HTML editor context if it is not expected to be generally used.

 Generally we avoid having any knowledge of orgUnitIds as everything works through hypermedia resources which avoid sharing concepts like the orgUnitId variable in the UI code. So I'm not sure if we have any other direct reference to the orgUnitId. But we seem to keep running up against these impedance mismatches between the needs of legacy and the nature of hypermedia.
 
 Which is kind of why I prefer ducking the whole problem and using the slug approach to embed the orgUnitId in with the collection item id. The fact that we are even having to directly reference the collection item id is a bit of a code smell, but again - legacy mismatches.
 
 So that would be my preferred approach but recognize it has much more impact.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -30,16 +32,50 @@ const componentClass = class extends HypermediaStateMixin(ListItemMixin(LitEleme
...
+
+	_openDialog() {
+		const open = D2L.LP.Web.UI.Legacy.MasterPages.Dialog.OpenFullscreen;
+		const url = new D2L.LP.Web.Http.UrlLocation(
+			`/d2l/lms/question/edit/${this.key}`);
+		return open(url, 'SrcCallBack', 'result', [], false, '');
+	}
+
+	_renderPrimaryAction(contentId) {
+		return html `<a aria-labelledby="${contentId}" href="#" @click="${this._handleLinkClick}"></a>`;
```
mdgbayly, line 76: nit: not sure about a11y implications of using an anchor tag in this way?

mdgbayly (approved):
Looks good - just the a11y concern with using the anchor tag. Maybe that can be reviewed and followed up in a subsequent PR if necessary.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
stevenao: the same mechanism is being used in d2l-list in learning paths. so, I am wondering how a11y is working out there... will find out later..

PR #150, opened 2021-03-23
US126109 dispatched update points event after collection refreshed

me-louie:
Dispatching `update-total-quiz-points` event from `d2l-activity-collection-item-quiz`. This event is caught and handled in `d2l-activity-quiz-editor` in the `activities` repo.

**Related PRs:**
`siren-sdk`: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/301
`activities`: https://github.com/BrightspaceHypermediaComponents/activities/pull/1624

[US126109](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F512322551332&fdp=true?fdp=true): Refresh total points on edit

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F512322551332&fdp=true?fdp=true

### DIFF
--- a/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
+++ b/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
@@ -60,7 +60,10 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
 		delayedResult.AddListener((/*result*/) => {
 			fetch(this._state, true).then(() => {
 				// refresh collection
-				fetch(Array.from(this._state._parents.keys())[0], true);
+				fetch(Array.from(this._state._parents.keys())[0], true).then(() => {
+					// refresh Quiz total points
+					this.dispatchEvent(new CustomEvent('update-total-quiz-points', {bubbles: true, composed: true}));
+				});
 			});
 		});
 	}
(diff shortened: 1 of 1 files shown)


components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -60,7 +60,10 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
 		delayedResult.AddListener((/*result*/) => {
 			fetch(this._state, true).then(() => {
 				// refresh collection
-				fetch(Array.from(this._state._parents.keys())[0], true);
+				fetch(Array.from(this._state._parents.keys())[0], true).then(() => {
+					// refresh Quiz total points
+					this.dispatchEvent(new CustomEvent('update-total-quiz-points', {bubbles: true, composed: true}));
```
mdgbayly, line 65: nit: we tend to name our events using the `d2l-` prefix. Also I wonder if we want to make this more generic given we don't know for sure it's points have changed?

We could possibly send an event that indicates a particular question/collection item was updated and include either/both the question activity id and the collection id. Or maybe just leave that out for now until we know for sure we need something like that. So maybe an event like `d2l-question-updated`?

me-louie: Thanks for the feedback, that makes sense. I will change it to `d2l-question-updated` for now.

mdgbayly (approved):

PR #151, opened 2021-03-24
Add myself as a codeowner of workToDo

bearfriend:


m6jones (approved):

m6jones:
It be great for you to ask questions on what I am doing and why. :)

m6jones:
I actually thought about it a bit more. I would prefer that you were not a code owner until you learned more about foundation engine. Since it is a new and emerging pattern and the documentation is still in the works I think it is for the best. I'll close it this for now and when I have it finished, I can show you the ropes and at that point I'll make you an owner.

bearfriend:
That's fair. Really just looking for a way to be kept abreast of things without pestering you and without getting every change to every component. I wish you could watch directories 😢 .

PR #162, opened 2021-03-30
US126257 refresh Section name on edit

me-louie:
Refreshing the Section name after editing using the new `_refreshState` in `foundation-engine`.
Not to be merged until https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/91 is merged.

[US126257](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F514848992408%2Ftasks&fdp=true?fdp=true): Edit a section

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F514848992408%2Ftasks&fdp=true?fdp=true

### DIFF
--- a/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
+++ b/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
@@ -4,7 +4,6 @@ import '../../../../list/custom/quiz/d2l-activity-list-item-section.js';
 import { css, LitElement } from 'lit-element/lit-element.js';
 import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
 import { fetch } from '@brightspace-hmc/foundation-engine/state/fetch.js';
-import { getComposedChildren } from '@brightspace-ui/core/helpers/dom.js';
 import { guard } from 'lit-html/directives/guard';
 import { html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
 
@@ -78,9 +77,7 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
 	}
 
 	_refreshSection() {
-		const children = getComposedChildren(this);
-		if (!children) return;
-		const section = children[0].getElementsByTagName('d2l-activity-list-item-section')[0];
+		const section = this.renderRoot.children[0].querySelector('d2l-activity-list-item-section');
 		if (!section) return;
 		section.refetch();
 	}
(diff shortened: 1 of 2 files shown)


stevenao (approved):

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -75,6 +76,12 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
 		return open(url, 'SrcCallBack', 'result', [], false, '');
 	}
 
+	_refreshSection() {
+		const section = this.renderRoot.children[0].querySelector('d2l-activity-list-item-section');
```
mdgbayly, line 80: This works but isn't particularly elegant but I recognize there isn't an obvious way to resolve dependencies like this between the parent and the child. But having a hard coded reference to the concrete child type `'d2l-activity-list-item-section' in this container element is not ideal. Let me ruminate on it for a short while to see if I can think of alternative ways of signalling to the child element.

me-louie: Thanks - yes I didn't love it either but I was having a hard time getting around it with the shadow root. Let me know if you come up with anything better.

mdgbayly: One option could be to use events to allow the child component to register an interest in knowing when things are changed which is how we wire up our different editors in FACE.

https://github.com/BrightspaceHypermediaComponents/activities/blob/04dfd80325d0a094014718ed0a3250e3e5a75e66/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js#L106

But that feels like overkill here. 

A more pragmatic solution might just be to declare an attribute/property on `d2l-activity-list-item-quiz` that can be set when an element has changed and then can be observed in the child element and used to trigger the refetch.

e.g. `d2l-activity-collection-item-quiz` has a new property `refreshSlug` or `refreshCounter` that can be set to some kind of value that can be modified each time an item is edited. e.g. a guid, or an integer counter.

Then in the dialog event handler you just change/increment that property. That will cause a render and pass the updated value of that property to the child component. In the child component you can either add an attribute change handler or a property setter to observe changes to that property and call the refetch.

```
		delayedResult.AddListener((/*result*/) => {
		    ...
		    ++this._refreshCounter;
		});
		
		render() {
		    ....
		    <d2l-activity-list-item-quiz number="${this.number}" href="${this._activityHref}"
				.token="${this.token}" points="${this._points}" refresh-counter="${this._refreshCounter}">
		}
```

This then keeps things decoupled and could be used by questions as well if needed (which we don't need right now).

mdgbayly: Still kind of ugly to have this hard wiring but this feels a little less invasive.

stevenao: We will need to have a way to skip list-item-question refresh that way b/c the way it requires a new URL. With that in mind, should we go with the "register editor" approach instead?

stevenao (approved):

PR #164, opened 2021-04-01
fix: Removing the discover rules label

zinabat:
## Context
[US120165](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F424331214124&fdp=true?fdp=true)

Removing the label for Discover rules because it's less complicated than hiding the existing one in the LMS. This should also ensure more consistent spacing, plus makes this component more reusable.

## Quality
- Demo page


BhamelinD2L (approved):
Incredibly risky PR but I trust you

PR #166, opened 2021-04-01
feat: Show categories and their items.

m6jones:
Make the parent list.
image

Basically categorizes what we are getting from the API and list what we are doing.

### DIFF
--- a/features/workToDo/d2l-w2d-list.js
+++ b/features/workToDo/d2l-w2d-list.js
@@ -1,0 +1,41 @@
+import '@brightspace-ui/core/components/list/list.js';
+import '@brightspace-ui/core/components/list/list-item.js';
+import '@brightspace-ui/core/components/list/list-item-content.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { W2dDateCategory } from './w2dDateCategoryObserver.js';
+
+class W2dList extends HypermediaStateMixin(LitElement) {
+	static get properties() {
+		return {
+			category: { type: String },
+			_activities: {
+				type: Array,
+				observable: observableTypes.custom,
+				observableObject: W2dDateCategory,
+				category: 'category'
+			}
+		};
+	}
+
+	constructor() {
+		super();
+		this._activities = [];
+		this.requiredPropertyForState('category');
+	}
+
+	render() {
+		return html`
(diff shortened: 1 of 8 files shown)


features/workToDo/d2l-w2d-list.js
```diff
@@ -0,0 +1,41 @@
...
+				observableObject: W2dDateCategory,
+				category: 'category'
+			}
+		};
+	}
+
+	constructor() {
+		super();
+		this._activities = [];
+		this.requiredPropertyForState('category');
```
lloydwaddell, line 24: what does this fancy magic do?

m6jones: Ohh! This fancy magic, basically says, wait until the value is set before making and getting the state. Since what we get from the state depends on it.

It also, will update the state if it changes. So if that attribute changes the component will update how the information should be given to it by the observable.

features/workToDo/w2dDateCategoryObserver.js
```diff
@@ -0,0 +1,72 @@
...
+
+const msInADay = 86400000;
+
+export class W2dDateCategory extends SirenSubEntities {
+
+	static definedProperty({ token, groupByDays, startDate, category, verbose }) {
+		return { token, groupByDays, startDate, category, verbose };
+	}
+
+	constructor({ id, token, state, verbose }) {
```
lloydwaddell, line 11: can this be removed?

features/workToDo/d2l-w2d-list.js
```diff
@@ -0,0 +1,41 @@
...
+import '@brightspace-ui/core/components/list/list-item.js';
+import '@brightspace-ui/core/components/list/list-item-content.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { W2dDateCategory } from './w2dDateCategoryObserver.js';
+
+class W2dList extends HypermediaStateMixin(LitElement) {
+	static get properties() {
+		return {
+			category: { type: String },
```
lloydwaddell, line 11: I don't see where this is used? is it a future implementation?

lloydwaddell (approved):

features/workToDo/d2l-w2d-list.js
m6jones: It is being used in the next property. you will see `category: "category"` and that is linking that value.

I am thinking that we should make that more obvious but that can come later.

PR #171, opened 2021-04-08
Awang/us126118 animation directive rule picker

AndyWangD2L:
[US126118](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F512393722300)

This adds an animation to the rule picker when new conditions are added/removed.


features/discover/d2l-discover-rule-picker.js
```diff
@@ -114,11 +120,12 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
 		}
 	}
 
-	_addDefaultCondition() {
+	_addCondition(state = 'new') {
```
zinabat, line 123: Can someone explain the `state` stuff to me? Are we just using these "states" to determine which animation to use or..?
These should probably be defined at the top in a frozen object so that it's very clear. IE:
```js
const conditionStates = Object.freeze({
    new: 'new',
    existing: 'existing',
    ...
});
```

features/discover/d2l-discover-rule-picker.js
```diff
@@ -25,15 +26,18 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
 	static get styles() {
 		return [ bodyCompactStyles, selectStyles,
 			css`
+				.relative-parent {
```
zinabat, line 29: nit: Classes should always be prefixed with `d2l-`

zinabat (commented):
Tests are probably failing because the removal itself happens after the animation is finished, so we'll need to wait for that event in the tests too.

features/discover/d2l-discover-rule-picker.js
AndyWangD2L: > Are we just using these "states" to determine which animation to use or..?
Yeah, it's pretty much for that. Will do that in a bit once I figure out the bug behind the tests failing

features/discover/d2l-discover-rule-picker.js
```diff
@@ -168,57 +187,86 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
...
+					this.conditions[index].properties.state = conditionStates.new;
+					this.requestUpdate();
+				}
+
+				this.dispatchEvent(new CustomEvent('d2l-rule-condition-removed', {
+					bubbles: true,
+					composed: true
+				}));
+			};
+			elem.addEventListener('d2l-animate-complete', onAnimateEnd);
```
dlockhart, line 212: Could this be added right inline in the `render()` via `.animate="${animateAction} @d2l-animate-complete="${this._handleAnimateComplete}"`? That way you don't need to worry about adding/remove the event listener on a parent node.

AndyWangD2L: Good idea, I added this in. However, this means we need to differentiate between whether `@d2l-animate-complete` triggers for a show or a hide.

zinabat (approved):

PR #175, opened 2021-04-13
feat: Style Work to Do

m6jones:
image
image


tenbradley:
This might be out of scope for the story, but it seems like only one of the empty widget things is appearing.
Aside from that, looks good.

tenbradley (approved):
Looking good and fancy.

PR #177, opened 2021-04-14
fix: empty question name on collection refetch

stevenao:


me-louie (approved):
Cool.. so the issue was with how we were refreshing the collection?

PR #183, opened 2021-04-16
fix quiz collection item checkbox at root level

me-louie:
Fix for:  "When using FACE Quiz, user cannot check the checkmark next to an existing question, section, (assume question pool too). Instead it goes into edit.  Seems like the click target is the whole row?"

Note this bug still happens for items nested within sections but the proper way to render nested lists are still a WIP.

https://trello.com/c/GKSEmBoY

### DIFF
--- a/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
+++ b/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
@@ -85,18 +85,6 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
 			`/d2l/lms/question/edit/${this.key}`);
 		return open(url, 'SrcCallBack', 'result', [], false, '');
 	}
-	_renderCheckbox() {
-		const disabled = this.disabled || this.skeleton;
-		return this.selectable ? html`
-			<input
-				id="${this._checkboxId}"
-				class="d2l-input-checkbox d2l-skeletize"
-				@change="${this._onCheckboxChange}"
-				type="checkbox"
-				.checked="${this.selected}"
-				?disabled="${disabled}">
-			` : html ``;
-	}
 	_renderPrimaryAction(contentId) {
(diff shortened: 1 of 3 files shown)


components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -78,7 +79,19 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
...
 	}
-
+	_renderCheckbox() {
+		const disabled = this.disabled || this.skeleton;
+		return this.selectable ? html`
+			<input
+				id="${this._checkboxId}"
+				class="d2l-input-checkbox d2l-skeletize"
+				@change="${this._onCheckboxChange}"
+				style="margin-top: 0.7rem"
```
me-louie, line 89: I tried to do this in `get styles()` i.e. 		
```
.d2l-input-checkbox {
			margin-top: 0.7rem;
		}
```
but it was getting overridden

stevenao (approved):

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -33,13 +33,20 @@ const componentClass = class extends HypermediaStateMixin(ListItemLinkMixin(LitE
 	}
 
 	static get styles() {
-		return [ super.styles, css `` ];
+		return [
+			super.styles,
+			css `
+				input[type="checkbox"].d2l-input-checkbox {
```
m6jones, line 39: If using the native checkbox for the list. I would push back on editing the styles if you can. Since we want to be consistent.

m6jones:
So, for lists the contain list. We will probably need a different type of list item. The list item will have one main click target and that is for accessibility. 

Using list checkbox also makes sure the click area for them is the biggest it can be. 

Another way to get around it is to remove content action if you have a sub list.

When using lists they are just special formated, ul and li tags. So don't do anything you wouldn't do with those.

m6jones:
Also, it looks like you want to extend listitembuttonmixin which will give you the proper actions

PR #186, opened 2021-04-22
fix: move start date to the attribute list

tenbradley:
### Context:
[US127302](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F600483964584&fdp=true?fdp=true)
The start date in full screen is currently displayed in line with the description. We want it to be inline with the type of activity and course.

### DIFF
--- a/features/workToDo/d2l-w2d-list-item.js
+++ b/features/workToDo/d2l-w2d-list-item.js
@@ -121,7 +121,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 
 		const startDate = (!this.actionHref && this._dates.start)
 			? html`
-				<d2l-status-indicator slot="supporting-info" state="none" text="Starts ${formatDate(this._dates.start, {format: 'shortMonthDay'})}"></d2l-status-indicator>
+				<div><d2l-status-indicator slot="supporting-info" state="none" text="Starts ${formatDate(this._dates.start, {format: 'shortMonthDay'})}"></d2l-status-indicator></div>
 			`
 			: nothing;
 
@@ -131,9 +131,8 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 				<d2l-list-item-content>
 					<d2l-activity-name class="d2l-w2d-list-item-name" href="${this.href}" .token="${this.token}"></d2l-activity-name>
 					<d2l-w2d-attribute-list slot="secondary">
-						${this._renderAttributeListCollapsed()}
+						${startDate} ${this._renderAttributeListCollapsed()}
 					</d2l-w2d-attribute-list>
-					${startDate}
 					${ !this.collapsed ? html`<d2l-activity-description slot="supporting-info" href="${this.href}" .token="${this.token}"></d2l-activity-description>` : nothing}
 				</d2l-list-item-content>
 			`)}`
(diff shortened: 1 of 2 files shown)


tenbradley:
w2d-fullscreen-startdate
Full screen start date.

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -133,10 +133,9 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 			content: html`${guard([this.href, this.token], () => html`
 				<d2l-list-item-content>
 					<d2l-activity-name class="d2l-w2d-list-item-name" href="${this.href}" .token="${this.token}"></d2l-activity-name>
-					<d2l-w2d-attribute-list slot="secondary" >
-						${this._renderAttributeListCollapsed()}
+					<d2l-w2d-attribute-list slot="secondary">
+						${startDate} ${this._renderAttributeListCollapsed()}
 					</d2l-w2d-attribute-list>
-					${startDate}
```
m6jones, line 139: Start date should be here when it is collapsed

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -133,10 +133,9 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 			content: html`${guard([this.href, this.token], () => html`
 				<d2l-list-item-content>
 					<d2l-activity-name class="d2l-w2d-list-item-name" href="${this.href}" .token="${this.token}"></d2l-activity-name>
-					<d2l-w2d-attribute-list slot="secondary" >
-						${this._renderAttributeListCollapsed()}
+					<d2l-w2d-attribute-list slot="secondary">
+						${startDate} ${this._renderAttributeListCollapsed()}
```
m6jones, line 137: If it isn't collapsed then it moves up here.

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -124,7 +124,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 
 		const startDate = (!this.actionHref && this._dates.start)
 			? html`
-				<d2l-status-indicator slot="supporting-info" state="none" text="Starts ${formatDate(this._dates.start, {format: 'shortMonthDay'})}"></d2l-status-indicator>
+				<div><d2l-status-indicator slot="supporting-info" state="none" text="Starts ${formatDate(this._dates.start, {format: 'shortMonthDay'})}"></d2l-status-indicator></div>
```
m6jones, line 127: The reason you have to wrap this in a div is because of the `slot` attribute. 
It would be cool if you only add the `supporting-info` slot if it is collapsed.

tenbradley: I added the div because I figured we wanted a dot between the start date and activity type. But we don't want a dot. Is there some other reason to have the div?

m6jones: If we don't want the dot. Then the best way to fix that is we can just change the slot based on collapsed or not.

m6jones: div shouldn't provide dotage.

m6jones (approved):

PR #190, opened 2021-04-23
US126706 maintain Map of items to be deleted

me-louie:
First draft implementation for quiz collection item delete.
Still need to look into `foundation-engine` allowing 204 - No Content response and refreshing the UI/adding link header.

[US126706](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F519271058560&fdp=true?fdp=true): [ui] delete questions and question pools from a root level


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -60,18 +65,21 @@ class ActivityEditorMainCollection extends HypermediaStateMixin(LitElement) {
...
 		this.addEventListener('d2l-question-updated', this._handleQuestionUpdate);
+		this.addEventListener('d2l-delete-quiz-collection-items', this._deleteCurrentSelection);
 	}
 
 	render() {
 		return html`
 			<div class="d2l-activity-collection-body">
 				<div class="d2l-activity-collection-body-content">
 					<div class="d2l-activity-collection-list-actions">
+						<d2l-activity-collection-item-delete-quiz ?skeleton="${this.skeleton}"></d2l-activity-collection-item-delete-quiz>
```
zinabat, line 78: You need to pass an href and token since this is a hypermedia component

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-delete-quiz.js
```diff
@@ -0,0 +1,65 @@
...
+            </div>
+		`;
+	}
+	_handleDialogClose() {
+		this._dialogOpened = false;
+	}
+	_handleDialogOpen() {
+		this._dialogOpened = true;
+	}
+	_onDeleteSelectedContent() {
```
zinabat, line 61: Unless you actually need this event, you can just directly commit the delete action here and update the state. You'll need to observe the quiz items from this component. You can see where we're doing this in the learning paths component:
https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/components/activity/editor/collection/d2l-activity-editor-main-collection.js#L113

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-delete-quiz.js
```diff
@@ -0,0 +1,65 @@
...
+import { html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
+import { HypermediaStateMixin } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { LocalizeCollectionAdd } from '../../lang/localize-collection-add.js';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+
+class ActivityCollectionItemDeleteQuiz extends SkeletonMixin(RtlMixin(HypermediaStateMixin(LocalizeCollectionAdd(LitElement)))) {
+	static get properties() {
+		return {
+			_dialogOpened: { type: Boolean },
```
zinabat, line 13: I imagine you will want to observe the collection items in this component so that you can alter their state

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-delete-quiz.js
me-louie: hi @zinabat , thanks for your feedback! So we should be declaring the delete action in this class, i.e.
```
			delete: {
				type: Object,
				observable: observableTypes.action,
				name: 'remove-activity',
				route: [{ observable: observableTypes.link, rel: rels.item }]
			}
```
and then on confirmation, commit each action, and then push/update the state here? In addition to observing the quiz items from this component, we should also be observing the Map of selected items which should be deleted?

zinabat: You'll get the delete in the `item`'s component instead of this one, and just call it from this controlling component instead rather than trying to route to it since there are more than one.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -26,18 +25,22 @@ class ActivityEditorMainCollection extends HypermediaStateMixin(LitElement) {
 				route: [
 					{ observable: observableTypes.link, rel: rels.collection }
 				]
-			}
+			},
+			_dialogOpened: { type: Boolean },
+			_selectedCandidates: { type: Object },
```
mdgbayly, line 30: Is this no longer used ? `_selectedCandidates`?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -89,6 +122,18 @@ class ActivityEditorMainCollection extends HypermediaStateMixin(LitElement) {
 		e.detail.reorder(this.items, { keyFn: (item) => item.properties.id });
 		this.requestUpdate('items', []);
 	}
+	_onDeleteSelectedContent() {
+		this._state.push();
```
mdgbayly, line 126: I don't see a `reset` on cancel now? Is that handled in some other way?

me-louie: Added reset on dialog Cancel

mdgbayly (approved):

PR #199, opened 2021-04-29
fix: lang terms are not being used for W2D

tenbradley:
### Context:
[DE43357](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F602042657008&fdp=true?fdp=true)
The lang terms that were taken from the activities repo have a different format than what we need for dynamic mixins. This fixes the translations to work with the mixin.

### Quality:
Tested with local build.


m6jones (approved):

bearfriend:
Just an FYI the dynamic mixin gives you control of what to import, so in `activities` we were importing `.val`. I mostly left it that way temporarily just as validation that it worked, but this (`default`) is ultimately what we want to encourage and use.

PR #202, opened 2021-04-29
Zina/rules list bug

zinabat:
This PR is to demonstrate a bug occurring that appears to be an issue with the engine's `updateProperties` function.

Kapture 2021-04-29 at 12 46 13

**To Reproduce**
1. The initial entity has subentities already attached to it. In this example, there is one.
2. Adding a second rule should add the rule to the list
3. Instead, the rule is added for a split second, then the list of rules reverts back to its original state
4. This **does not occur if the list starts as empty**
5. This also **does not occur if we reference the rules list in the top level component only and do not use the card component**

**Hypermedia Tree**
Course (BFF)
-- Entitlement
---- Rules (selfless)
------ Conditions (selfless)
---- Condition Types

**Component Tree**
`<d2l-discover-rules href="course-bff">` - listens to routed rules
-- `<d2l-discover-rules-dialog href="entitlement">` - listens and changes rules
---- `<d2l-discover-rule-picker href="entitlement">` - listens to rules
------ `<d2l-discover-attribute-picker href="condition-type">` - listens to condition-types
-- `<d2l-discover-rules-card href="entitlement">` - listens and changes rules


features/discover/d2l-discover-rule-picker.js
```diff
@@ -145,7 +145,7 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
 				type: this.defaultType || (this.conditionTypes && this.conditionTypes[0].properties.type),
 				values: [],
 				state: state
-			}
+			},
 		});
 		this.requestUpdate();
```
m6jones, line 150: this should be `this.requestUpdate('conditions', [])`

zinabat:
This issue has been solved via https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/101

PR #206, opened 2021-04-30
fix: safari bullets not showing up

m6jones:
Bullets don't show up in fullscreen and only fullscreen so this is probably the fix.

https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F601797704532


zinabat:
This does not work in Safari. Usage of pseudo elements after slotted is not supported in Safari https://bugs.webkit.org/show_bug.cgi?id=178237

I recommend adding the before and after from the controlling component, in the `list-item`, or you can add actual elements and hide the first one in the attribute list.

PR #211, opened 2021-05-06
fix: W2d total pages minimum of 1

tenbradley:
### Context:
Currently, the total number of pages may appear to be zero. This makes it so that the number can't be below 1.

### DIFF
--- a/features/workToDo/d2l-w2d-collections.js
+++ b/features/workToDo/d2l-w2d-collections.js
@@ -213,6 +213,7 @@ class W2dCollections extends LocalizeDynamicMixin(HypermediaStateMixin(LitElemen
 		this.collapsed = false;
 		this._totalActivities = 0;
 		this.__currentPageOverdue = 1;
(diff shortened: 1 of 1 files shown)


features/workToDo/d2l-w2d-collections.js
```diff
@@ -427,6 +427,9 @@ class W2dCollections extends LocalizeDynamicMixin(HypermediaStateMixin(LitElemen
 		if (this._pagingTotalResultsOverdue && !this._lastOverduePageHasMoreThanHalf()) {
 			totalPages -= 1;
 		}
+		if (totalPages < 0) {
```
lloydwaddell, line 430: what if totalPages is 0, should that be handled here too?

tenbradley: Darn it. :(

tenbradley: Fixed.

lloydwaddell (approved):

PR #222, opened 2021-05-14
Bhamelin/us126578 match count

BhamelinD2L:
[US126578 ](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F517443460728)- Add Match-Count to the discover rule picker

This PR does a few different things:
- Creates and utilizes `match-count-mixin` in the Discover Rule Picker (and eventually the Discover Rule Card) to connect to the BFF and retrieve the Match Count via a summon action on attribute change/rule change/ attribute delete
- Edits some of the events to make the dialog resize occur on attribute add, and after the dialog loads new attributes. This should prevent the rule count from being scrolled out of view by virtue of creating a new attribute.
 
- Updates `es-dev-server` to `web-dev-server` - This appears to have caused no issues for any of the demos, which is good as `es-dev-server` recommends this switch in its readme. The Learning Path demo does not work for me on either server, so if it's supposed to, that should be investigated.

- Uses `web-dev-server` with `koa-proxies` to create a POST server to simulate a BFF response for match-count in its demos.

Match Count


features/discover/d2l-discover-rule-picker.js
```diff
@@ -144,6 +157,7 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
 		this.conditions.push({
 			properties: {
 				type: this.defaultType || (this.conditionTypes && this.conditionTypes[0].properties.type),
+				id: this._conditionTypesHash[this.defaultType] || (this.conditionTypes && this.conditionTypes[0].properties.id),
```
BhamelinD2L, line 160: I added infrastructure to conditions to track their `id` values across the components, as the id is what needs to be sent to the bff to determine match count here as well as in the card.

features/discover/d2l-discover-rule-picker.js
```diff
@@ -297,12 +316,28 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
...
 					state: conditionStates.existing,
 					values: [...condition.properties.values]
 				}
 			};
 		});
+		await this.requestUpdate();
+		this._sizeChanged();
+	}
+
+	_sizeChanged() {
```
BhamelinD2L, line 330: I added general function for requesting size changes. We were resizing on adding a new condition, and I think we also want to resize whenever the attribute picker's list wraps around to a newline, as well as once all the attributes are fully loaded in, as they influence the total height. 
A user can still get scrolling before full dialog size if they highlight a very long attribute list as the attribute [x] buttons appearing makes the list bigger, but if they modify the attribute list in there it will then resize to accommodate again.

features/discover/mixins/match-count-mixin.js
```diff
@@ -0,0 +1,39 @@
...
+			}
+		});
+		return {
+			match: matchArray,
+			...(includeUsers === true && { includeUsers: includeUsers }),
+			...(userLimit && { limit: userLimit })
+		};
+	}
+
+	async getMatchData(summonAction, conditions, includeUsers, userLimit) {
```
BhamelinD2L, line 25: The goal here is that both the card and rule picker can call this to get back the data they need without duping the filtering logic

web-dev-server.config.mjs
```diff
@@ -0,0 +1,18 @@
+import http from 'http';
+import proxy from 'koa-proxies';
+import { handleRequest as handleDiscoverRequests} from './features/discover/demo/discover-api.js';
+
+const server = http.createServer((request, response) => {
+	handleDiscoverRequests(request, response);
+});
+
```
BhamelinD2L, line 8: I wanted this to be as straightforward as possible to use without tinkering with it for too long - If any other components want to mock a POST it should be as simple as importing your request handler from your feature and calling it in this function.

zinabat: u my hero

features/discover/d2l-discover-rule-picker.js
```diff
@@ -24,9 +25,8 @@ const conditionStates = Object.freeze({
 	remove: 'remove',
 	removed: 'removed'
 });
-
-class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitElement))) {
-
+// todo: edit an existing rule
```
zinabat, line 28: This is no longer a todo

features/discover/d2l-discover-rule-picker.js
```diff
@@ -36,6 +36,8 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
 			] },
 			_rules: { type: Array, observable: observableTypes.subEntities, rel: rels.rule },
 			_defaultType: { type: String },
+			_matchCount: { type: Number, rel: rels.matchCount },
```
zinabat, line 39: looks like the `rels.matchCount` got left in here by accident?

features/discover/test/d2l-discover-rule-picker.test.js
```diff
@@ -187,6 +188,55 @@ describe('d2l-discover-rule-picker', () => {
...
+			el._rules[3]  = { entities: [
+				{ properties: { id:'_fruit', state: 'existing', type: 'Fruit', values: [] } },
+				{ properties: { id:'_entree', state: 'existing', type: 'Entree', values: ['Spaghetti'] } }
+			] };
+			el._rules[4]  = { entities: [
+				{ properties: { id:'_fruit', state: 'existing', type: 'Fruit', values: [] } },
+				{ properties: { id:'_entree', state: 'existing', type: 'Entree', values: [] } }
+			] };
+
+			const matchCountDiv = el.shadowRoot.querySelector('#match-count');
```
zinabat, line 211: You should be able to target the class here

features/discover/d2l-discover-rule-picker.js
```diff
@@ -109,7 +115,10 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
 					@click="${this._addDefaultCondition}"></d2l-button-subtle>
 				<div class="d2l-picker-hr-match-separator">
 					<div class="d2l-picker-hr"></div>
-					<div class="d2l-body-compact">${this.localize('text-rule-matches', 'count', 'xxx')}</div>
+						<div id="match-count" class="d2l-picker-rule-match-count d2l-body-compact">
```
zinabat, line 118: nit: I usually try to avoid adding `id`s that are specific to tests if I can help it

features/discover/d2l-discover-rule-picker.js
```diff
@@ -297,12 +316,28 @@ class RulePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin(LitE
...
 			return {
 				properties: {
+					id: condition.properties.id,
 					type: condition.properties.type,
 					state: conditionStates.existing,
 					values: [...condition.properties.values]
 				}
 			};
 		});
+		await this.requestUpdate();
```
zinabat, line 326: Do we need to explicitly ask for an update since `conditions` has changed?

features/discover/d2l-discover-rule-picker.js
zinabat: Can we change all of them to use this function to launch the event? I don't think `d2l-rule-condition-added` is used for anything except size updates right now, but idk

features/discover/demo/discover-api.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-env node */
+const fs = require('fs');
+module.exports.handleRequest = function(request, response) {
+
```
zinabat, line 4: yay!

package.json
```diff
@@ -10,7 +10,7 @@
     "lint": "npm run lint:eslint && npm run lint:lit",
     "lint:eslint": "eslint . --ext .js,.html",
     "lint:lit": "lit-analyzer  \"{components,demo,features,test}/**/*.js\" --strict --rules.no-unknown-tag-name off --rules.no-incompatible-type-binding off --rules.no-unknown-attribute off --rules.no-unknown-property off",
-    "start": "es-dev-server --node-resolve --dedupe --open --watch",
+    "start": "web-dev-server --node-resolve --dedupe --open --watch",
```
zinabat, line 13: yay!!!

zinabat (commented):
Looking good! Just a couple minor things

features/discover/d2l-discover-rule-picker.js
BhamelinD2L: I think it was necessary at first, but I rewrote the structure of the match count for a different reason and should have cleaned it up. It's been done, thanks for pointing it out.

features/discover/d2l-discover-rule-picker.js
BhamelinD2L: Awaiting updateComplete instead is sufficient here, thanks.

features/discover/d2l-discover-rule-picker.js
BhamelinD2L: Oh, strange, maybe a rebase pulled that in 🤔

zinabat (approved):

PR #224, opened 2021-05-20
fix: Fix commonjs import

zinabat:
## Context
I ran into an issue with @BhamelinD2L 's change to `web-dev-server` that prevented me from running demos. I fixed it.
Screen Shot 2021-05-20 at 12 07 44 PM

## Quality
`rm -rf node_modules`
`rm package-lock.json`
`npm install`
`npm start`

Works now instead of erroring.

### DIFF
--- a/web-dev-server.config.mjs
+++ b/web-dev-server.config.mjs
@@ -1,4 +1,4 @@
-import handleDiscoverRequests from './features/discover/demo/discover-api.js';
+import { handleDiscoverRequests } from './features/discover/demo/discover-api.mjs';
 import http from 'http';
 import proxy from 'koa-proxies';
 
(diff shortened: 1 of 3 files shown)


web-dev-server.config.mjs
```diff
@@ -1,18 +1,18 @@
+import handleRequest from './features/discover/demo/discover-api.js';
 import http from 'http';
 import proxy from 'koa-proxies';
-import { handleRequest as handleDiscoverRequests} from './features/discover/demo/discover-api.js';
 
 const server = http.createServer((request, response) => {
-	handleDiscoverRequests(request, response);
+	handleRequest(request, response);
```
BhamelinD2L, line 6: Can we rename this function in `discover-api.js` to handleDiscoverRequest? I don't want to reserve `handleRequest` for one feature

BhamelinD2L (approved):

PR #228, opened 2021-05-27
fix: version bump

lloydwaddell:
trigger release for translations


tenbradley (approved):
Not sure. This seems like a big scary change.

PR #230, opened 2021-06-04
fix: Use composed event for discover rules

zinabat:
## Context
The checkbox portion of the component in the LMS is not behaving as expected because the current event it is attempting to catch, `d2l-checkbox-drawer-checked-change`, is not `composed`. Composed events are able to pierce the shadow DOM.

- Adds a new bubbling, composed event: `d2l-rules-checkbox-change`

## Quality
- New unit test added


m6jones (approved):
That they do

PR #232, opened 2021-06-04
fix: US128809 UI improvements

alogan519:


richmar10 (approved):

features/lti/d2l-lti-activity.js
```diff
@@ -98,28 +131,34 @@ class LtiActivity extends SkeletonMixin(LocalizeDynamicMixin(HypermediaStateMixi
...
 
 		html`${this._showOpenInNewWindowButton ?
 			html`<d2l-button class="spanning-button" primary @click="${this._onOpenInNewWindowClick}">${this.localize('open-in-new-window')}</d2l-button>` :
-			html`<iframe allow="microphone *; camera *; autoplay *" width="${this.iFrameWidth}px" height="${this.iFrameHeight}px" src="${this._launchUrl}"></iframe>
-						<div>
-							<d2l-button-subtle text="${this.localize('open-in-new-window')}" icon="tier1:new-window" @click="${this._onOpenInNewWindowClick}"></d2l-button-subtle>
-						</div>`
+			html`
+				<div class="content-frame">
+					<iframe allow="microphone *; camera *; autoplay *" width="${this.iFrameWidth}px" height="${this.iFrameHeight}px" src="${this._launchUrl}"></iframe>
```
maggiesazio, line 154: Do we want to set a bigger default for the iframe if we don't get height/width? I'm fine if we do that separately but I think we should probably do it.

maggiesazio (approved):

features/lti/d2l-lti-activity.js
alogan519: Sure ya I can do that separate, good idea.

sammeyerreed (approved):

PR #237, opened 2021-06-08
fix: send an encoded version of assigned-activity-href instead of just true

alogan519:


maggiesazio (approved):
Does the readme for the component need an update or is it ok?

alogan519:
> Does the readme for the component need an update or is it ok?

Updated it, thanks.

richmar10 (approved):

PR #238, opened 2021-06-08
US128875 Style FACE quiz editor questions when RHP is collapsed

dustxd:
### Motivation/Context
RHP is now collapsible to allow questions benefit from the extra space so we need to remove the width cap on it

### Summary of Changes
* Removed max width for both delete question button and questions

### Rally Story
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F601091699237&fdp=true?fdp=true


stevenao:
With the RHP not collapsed, just make sure long Question text gets wrapped probably.

dustxd:
This is how it looks like
image

wongvincent (approved):

PR #244, opened 2021-06-14
US128875 Remove left margin for RTL

dustxd:
### Motivation/Context
Questions within sections have a left margin that prevents RTL from rendering properly
image

### Summary of Changes
Removed left margin as it does not make a visual difference in normal LTR languages

_LTR_
image

_RTL_
image

### Rally Story
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F601091699237&fdp=true?fdp=true


wongvincent (approved):
Your change is fine, but line 83 needs a reverse `margin-right` for RTL

PR #245, opened 2021-06-14
Update activities to use the LabelMixin

dbatiste:
This PR updates activities to extend the core `LabelMixin` so that these components can be referenced as labels for list-items and other future selectable components such as in tables. 

I initially went down the path of the factoring the contents of `d2l-hc-name` into a mixin in order to "flatten" the components and enable the `updateLabel()` call, however the Lti case is more than just a name, and in such cases it "has a" name, rather than "is a" name, so extending a name mixin doesn't make sense.

Ref: https://github.com/BrightspaceUI/core/pull/1399

### DIFF
--- a/components/common/d2l-hc-name.js
+++ b/components/common/d2l-hc-name.js
@@ -32,6 +32,16 @@ class HmName extends SkeletonMixin(HypermediaStateMixin(LocalizeCommon(LitElemen
 		return html`<span class="d2l-skeletize">${this.name ? this.name : html`${this.localize('name')} <div class="d2l-activity-name-skeleton-extend-skeleton-width"></div>`}</span>`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (!changedProperties.has('name')) return;
+		this.dispatchEvent(new CustomEvent('d2l-label-change', {
+			bubbles: true,
+			composed: true,
+			detail: this.name
+		}));
+	}
+
 	get _loaded() {
 		return !this.skeleton;
 	}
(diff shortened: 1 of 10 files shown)


components/common/d2l-hc-name.js
```diff
@@ -32,6 +32,16 @@ class HmName extends SkeletonMixin(HypermediaStateMixin(LocalizeCommon(LitElemen
 		return html`<span class="d2l-skeletize">${this.name ? this.name : html`${this.localize('name')} <div class="d2l-activity-name-skeleton-extend-skeleton-width"></div>`}</span>`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (!changedProperties.has('name')) return;
+		this.dispatchEvent(new CustomEvent('d2l-label-change', {
```
dbatiste, line 38: The core `LabelMixin` will be updated to listen for this event.

m6jones: I am not a big fan of sending browser events with api data. This tends to leads to consistence issues if the api changes quickly.

zinabat: Can you think of an alternative @m6jones ? I agree with you in most cases regarding events, but at a certain point we need to be able to signal that the label has changed and needs to be updated at a higher level in the DOM

dbatiste: One alternative that Zina and I discussed was that [d2l-activity-editor-collection-add](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/components/activity/editor/collection/d2l-activity-editor-collection-add.js#L140) be refactored so that explicit `label` properties can be specified for the `d2l-list-item`s... that is the source issue that is driving this.  However, that refactoring I think is beyond what I can do at the moment, and also, I'm not _really_ super keen on the idea that all such consumers of list-items would have to create custom list-items.  That was something we'd explicitly tried to avoid.  Maybe there is another way though.

m6jones: If we had to do that, we would need to follow this pattern or make it so we can have dynamic routes.
https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/workToDo/d2l-w2d-list-item.js

dbatiste: How often do we expect these values to change (and thus how often would we expect the event)?

zinabat: I honestly think the event approach is fine until we can get some time to refactor the list-items inside of the `collection-add` component, or implement dynamic routes

dbatiste: Are we "ok" to merge this?

zinabat (approved):

PR #253, opened 2021-06-21
fix: make the width 100% for the lti component and change the hc-name…

alogan519:
… component to no-wrap only when skeleton is set.

### DIFF
--- a/components/common/d2l-hc-name.js
+++ b/components/common/d2l-hc-name.js
@@ -12,14 +12,13 @@ class HmName extends SkeletonMixin(HypermediaStateMixin(LocalizeCommon(LitElemen
 
 	static get styles() {
 		return [ super.styles, css`
-
-			:host([skeleton]) {
-				overflow-y: hidden;
-				white-space: nowrap;
-			}
 			.d2l-activity-name-skeleton-extend-skeleton-width {
 				display: inline-block;
 				min-width: 5rem;
+			}
+			.d2l-skeletize {
+				overflow-y: hidden;
+				white-space: nowrap;
 			}
 		`];
 	}
--- a/components/common/d2l-hc-name.js
+++ b/components/common/d2l-hc-name.js
@@ -12,13 +12,14 @@ class HmName extends SkeletonMixin(HypermediaStateMixin(LocalizeCommon(LitElemen
 
 	static get styles() {
 		return [ super.styles, css`
+
+			:host([skeleton]) {
+				overflow-y: hidden;
+				white-space: nowrap;
+			}
 			.d2l-activity-name-skeleton-extend-skeleton-width {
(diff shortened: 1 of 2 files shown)


maggiesazio (approved):

components/common/d2l-hc-name.js
```diff
@@ -12,14 +12,15 @@ class HmName extends SkeletonMixin(HypermediaStateMixin(LocalizeCommon(LitElemen
 
 	static get styles() {
 		return [ super.styles, css`
+
+			:host([skeleton]) {
```
dbatiste, line 16: It looks like this change would affect a bunch of things.  I presume that's what we want?

features/lti/d2l-lti-activity.js
```diff
@@ -98,9 +99,16 @@ class LtiActivity extends SkeletonMixin(LocalizeDynamicMixin(LabelMixin(Hypermed
 			.header-name {
 				margin-bottom: 1.5rem;
 			}
+			.icon {
+				margin-left: 0.8rem;
```
dbatiste, line 103: This will require adding the `RtlMixin` and adding a corresponding rtl selector to flip this whitespace.

Looks like there is only the one icon in this component... so this selector could probably be changed to `d2l-icon { ... }` and the `.icon` class could be avoided.

zinabat (changes requested):
I would second Dave's RTLmixin suggestion before merging. Otherwise,

features/lti/d2l-lti-activity.js
```diff
@@ -163,7 +176,7 @@ class LtiActivity extends SkeletonMixin(LocalizeDynamicMixin(LabelMixin(Hypermed
 			html`<d2l-button class="spanning-button" primary @click="${this._onOpenInNewWindowClick}">${this.localize('open-in-new-window')}</d2l-button>` :
 			html`
 				<div class="content-frame">
-					<iframe allow="microphone *; camera *; autoplay *" width="${this.iFrameWidth}px" height="${this.iFrameHeight}px" src="${this._launchUrl}"></iframe>
+					<iframe class="${this.iFrameWidth ? '' : 'content-frame-default-width'}" allow="microphone *; camera *; autoplay *" width="${this.iFrameWidth}px" height="${this.iFrameHeight}px" src="${this._launchUrl}"></iframe>
```
zinabat, line 179: I'm approving this now, but normally when we add classes this way, we use the `classMap` directive: https://lit-html.polymer-project.org/guide/template-reference#classmap

zinabat (approved):

features/lti/d2l-lti-activity.js
alogan519: oh nice that's perfect, thank you

PR #261, opened 2021-06-29
US128094 Add start/end date observers

ryantmer:
As part of adding the new W2D component to Brightspace for Parents, we need to be able to change the date range W2D is showing, as this was a feature of the old upcoming work widget in BfP. W2D mostly works in this regard, but it doesn't fetch the new date range when an updated start/end date is set on it. This change resolves that by re-fetching whenever these parameters change.

This also includes a minor refactor by moving some common behaviour into `_fetchRoutedState`, as it was needed by two existing functions, as well as the two new ones.

### DIFF
--- a/features/workToDo/w2dSummonActionObserver.js
+++ b/features/workToDo/w2dSummonActionObserver.js
@@ -19,6 +19,8 @@ export class W2dSummonAction extends SirenSummonAction {
 		}
 		super.addObserver(observer, property, { method, route });
 		this._setPage(observer[page]);
+		this._setStartDate(observer[start]);
+		this._setEndDate(observer[end]);
 	}
 
 	get method() {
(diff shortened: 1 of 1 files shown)


ryantmer (commented):
My first foray into anything HMC-related, so hopefully this isn't the worst change ever created. Still a little fuzzy on exactly how this all works, but I figured repeating an existing pattern likely wasn't too risky!

features/workToDo/w2dSummonActionObserver.js
```diff
@@ -42,23 +44,39 @@ export class W2dSummonAction extends SirenSummonAction {
...
-			this._routes.forEach((route, observer) => {
-				this.routedState.addObservables(observer, route);
-			});
-			await fetch(this.routedState);
+			await this._fetchRoutedState();
+		}
+	}
+
+	async _fetchRoutedState() {
+		this.routedState = await this.createRoutedState(this.href, this._token.rawToken);
```
jstefaniuk-d2l, line 52: should the `if (this._routes.size > 0 && this._href && this._token) {` checks be moved here?

features/workToDo/w2dSummonActionObserver.js
```diff
@@ -42,23 +44,39 @@ export class W2dSummonAction extends SirenSummonAction {
...
+		this._routes.forEach((route, observer) => {
+			this.routedState.addObservables(observer, route);
+		});
+		await fetch(this.routedState);
+	}
+
+	async _setEndDate(endDate) {
+		if (!endDate || this._endDate === endDate) return;
```
jstefaniuk-d2l, line 60: any chance that "removing the end date" could be a case we would need to support?

lloydwaddell: heads up: the activity APIs that this is using require a start and end date.

ryantmer: Yeah, the API will return the empty state with neither start or end, but it returns a 400 if only one is specified. I wasn't seeing that come up, i.e. this only ever did requests that had both start and end set, but I think it's still technically possibly for a request to go out with just one or the other - I'll fix that.

features/workToDo/w2dSummonActionObserver.js
jstefaniuk-d2l: I think that it sets the href right above, so that check never got added, it does check that routes is there, and I think that if _token isn't there it'll cause an exception

ryantmer: Yeah, seems reasonable. Admittedly I didn't actually think any further on it 😅 But, you make a rational point, I'll combine those checks as well.

lloydwaddell (approved):

jstefaniuk-d2l (approved):
LGTM

PR #264, opened 2021-06-29
Hide "View All Work" button from empty state

ryantmer:
The "View All Work" button is being correctly hidden [in `d2l-w2d-collections`](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/65c3e3a7544907be18ed4a05b92b59a40af50667/features/workToDo/d2l-w2d-collections.js#L294) when the `dataFullPagePage` is not set. This just adds similar behavior to the empty state `d2l-w2d-no-activities` component, so that it doesn't display a "View All Work" button if there is nowhere for the button to go.

### DIFF
--- a/features/workToDo/d2l-w2d-no-activities.js
+++ b/features/workToDo/d2l-w2d-no-activities.js
@@ -155,7 +155,7 @@ class w2dNoActivities extends LocalizeDynamicMixin(HypermediaStateMixin(LitEleme
 	}
 
 	_renderEmptyViewButton() {
-		if (!this.activities) {
+		if (!this.activities || !this.dataFullPagePath) {
 			return undefined;
 		}
 		return html`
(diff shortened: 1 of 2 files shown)


features/workToDo/d2l-w2d-no-activities.js
```diff
@@ -155,7 +155,7 @@ class w2dNoActivities extends LocalizeDynamicMixin(HypermediaStateMixin(LitEleme
 	}
 
 	_renderEmptyViewButton() {
-		if (!this.activities) {
+		if (!this.activities || !this.dataFullPagePath || this.dataFullPagePath === 'undefined') {
```
ryantmer, line 158: Kind of an unfortunate fix, but this covers both `data-full-page-path=""` (which gets converted to just a `data-full-page-path` boolean attribute and results in `dataFullPagePath = false`, confusingly) and omitting the attribute entirely (which results in `dataFullPagePath = "undefined"`)

jstefaniuk-d2l (approved):

PR #267, opened 2021-07-09
Testing strategy for components

lloydwaddell:
```Context```
I had an idea of creating generic testing data that could be reused for components, but thinking this is more confusing / complex than it should be. In addition the AddMock util was only good for learning paths use case. Defining mocks inside the tests with the data seems to be a better approach, so I am updated the testTemplate and started to cleanup the tests.

Leaving the PR where it is for now to discuss if this is the approach we should follow.

### DIFF
--- a/components/activity/code/test/d2l-activity-code-editor.test.js
+++ b/components/activity/code/test/d2l-activity-code-editor.test.js
@@ -7,11 +7,11 @@ import fetchMock from 'fetch-mock/esm/client';
 import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
 import sinon from 'sinon/pkg/sinon-esm.js';
 
-fetchMock.config.overwriteRoutes = true;
-
 const inputText = 'd2l-input-text',
 	activityHref = 'http://activity/',
-	orgHref = 'http://org/';
+	activityHrefNoAction = 'http://activity/no-action',
+	orgHref = 'http://org/',
+	orgHrefNoAction = 'http://org/no-action';
 
 const activity = {
 	class:['activity'],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/testTemplate.js [hunk 1]
```diff
@@ -58,11 +61,11 @@ describe('component-name', () => {
 		});
 
 		afterEach(() => {
-			mockLink.resetHistory();
+			fetchMock.resetHistory();
 		});
 
 		it('should ...', async() => {
-			await _createComponent('path');
+			await _createComponent(apiHref);
```
AKobets, line 68: is this test needed? See similar test at line 46

components/activity/code/test/d2l-activity-code-editor.test.js
```diff
@@ -50,29 +80,18 @@ describe('d2l-activity-code-editor', async() => {
...
-		});
-
-		it('should initialize using defined path and expected values', async() => {
-			const element = await _createCodeEditor('/learning-path/new');
-
-			assert.equal(element.code, learningPathNew.properties.code);
-		});
-
-		describe('path:/learning-path/existing', () => {
+		describe('Code', () => {
```
AKobets, line 83: would it be better `describe('d2l-activity-code-editor',...`

components/activity/code/test/d2l-activity-code-editor.test.js
```diff
@@ -100,18 +119,20 @@ describe('d2l-activity-code-editor', async() => {
...
 		});
-		describe('path:/learning-path/missing-action', () => {
-			let element;
-			beforeEach(async() => {
-				clearStore();
-				element = await _createCodeEditor('learning-path/missing-action');
-				assert.equal(element.code, learningPathMissingAction.properties.code, 'code should match response');
-			});
+
+		describe('Update code action missing', () => {
```
AKobets, line 123: Does `code not updated when action missing` test needs its own describe section?

components/activity/code/test/d2l-activity-code-editor.test.js
```diff
@@ -100,18 +119,20 @@ describe('d2l-activity-code-editor', async() => {
...
+
 			it('code not updated when action missing', async() => {
+				fetchMock.mock(orgHref, JSON.stringify(orgNoAction));
+
+				clearStore();
```
AKobets, line 128: It looks like clearStore is already called at line 66

components/activity/code/test/d2l-activity-code-editor.test.js
lloydwaddell: I guess not lol

AKobets (approved):
Looks good.

PR #268, opened 2021-07-09
Bhamelin/us127861 user images

BhamelinD2L:
Adds user profile images and match count info to the rule cards.
[US127861](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F603051838404&fdp=true?fdp=true)
image

Adds `userList` to the rules entity model, and updates it with match count as a well as when the rules load.
Updates the JSON to include sample user info.
Adds a couple small tests for initializing the cards.


features/discover/d2l-discover-rule-card-match-info.js
```diff
@@ -0,0 +1,110 @@
...
+			.d2l-rule-card-profile-images-image {
+				width: 1rem;
+				border-radius: 0.3rem;
+				border: 0.1rem solid white;
+			}
+			.d2l-rule-card-profile-images-more {
+				width: 1rem;
+				line-height: 1.6rem;
+				text-align: center;
+			}
```
zinabat, line 39: nit: these should be ordered alphabetically

features/discover/d2l-discover-rules.js
```diff
@@ -203,7 +204,7 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
 	}
 
 	async _summonEntitlement() {
-		const sirenReponse = await this._getEntitlement.summon(null, true);
+		const sirenReponse = await this._getEntitlement.summon({profileCount : profileCount});
 		if (sirenReponse) {
 			const newRules = sirenReponse?.entities.filter(e => e.rel.includes(rels.rule));
```
zinabat, line 209:
```suggestion
			const newRules = sirenReponse.entities.filter(e => e.rel.includes(rels.rule));
```
You already have a check for existence wrapping it

zinabat (approved):
looks great - couple nits but they're optional

PR #269, opened 2021-07-12
Adding example components

lloydwaddell:
```Context```
Adding some example components for basic interactions with foundations-engine. Hoping to establish an area that can be used for learning foundations without having to create a new component from scratch.

is there any other interactions or known pitfalls that would be good to cover?

### DIFF
--- a/examples/d2l-hmc-example.js
+++ b/examples/d2l-hmc-example.js
@@ -1,0 +1,24 @@
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { customHypermediaElement } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
+
+export class HmcExample extends HypermediaStateMixin(LitElement) {
+
+	static get properties() {
+		return {
+			// Grab the `name` property from hypermedia response
+			name: { type: String, observable: observableTypes.property, name: 'name' },
+			// Get the element attribute
+			description: { type: String, attribute: 'description'}
+		};
+	}
+
+	render() {
+		return html`<div>
+			${this.name}: ${this.description}
+		</div>
+		`;
+	}
+}
+
+customHypermediaElement('d2l-hmc-example', HmcExample);
--- a/examples/demo/actions.json
+++ b/examples/demo/actions.json
@@ -1,0 +1,22 @@
+{
+  "actions": [
+    {
+      "href": "/update/title",
+      "name": "update-title",
+      "method": "PATCH",
+      "fields": [
+        {
+          "class": [
+            "required"
+          ],
+          "type": "text",
+          "name": "title",
+          "value": "default-title"
+        }
+      ]
+    }
+  ],
+  "properties": {
+    "title": "Example"
+  }
+}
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by AKobets** [comment]
Links to examples are missing on the demo page.
There is no reference to examples on the readme page.

File: examples/d2l-hmc-example.js [hunk 1]
```diff
@@ -0,0 +1,24 @@
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { customHypermediaElement } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
+
+export class HmcExample extends HypermediaStateMixin(LitElement) {
```
AKobets, line 5: should it be `HmcPropertiesExample`

examples/demo/d2l-hmc-examples.html
```diff
@@ -0,0 +1,36 @@
...
+		<title>d2l-hmc-example</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<meta charset="UTF-8">
+	</head>
+	<body>
+		<d2l-demo-page page-title="d2l-hmc-examples">
+			<h2>Properties and Attributes</h2>
+			<d2l-demo-snippet>
+				<d2l-hmc-example description="passing values by attributes" href="./name.json" token="cake"></d2l-hmc-example>
```
AKobets, line 21: Could you add name property observed from the ./name.json to be displayed in html ?

examples/demo/actions.json
```diff
@@ -0,0 +1,22 @@
...
+      "name": "update-title",
+      "method": "PATCH",
+      "fields": [
+        {
+          "class": [
+            "required"
+          ],
+          "type": "text",
+          "name": "title",
+          "value": "default-title"
```
AKobets, line 14: should it get title value from the hmc control input instead of hardcoded default value?

examples/d2l-hmc-example.js
lloydwaddell: renamed to HmcProperties to be consistent with other examples

examples/demo/d2l-hmc-examples.html
lloydwaddell: I could, but I want to draw a clear distinction between a property pulled from json and one from an attribute

examples/demo/actions.json
lloydwaddell: The component is observing the properties section. The default-title is only when saving a state without a value.

AKobets: > The component is observing the properties section. The default-title is only when saving a state without a value.

how then I've seen both titles in request inspector ? https://d2l.slack.com/archives/GUNAMLS92/p1626199305098900?thread_ts=1626102962.097200&cid=GUNAMLS92

lloydwaddell: I've confirmed the behaviour using learning paths, this seems to be a bug
hmc-engine-query-bug

lloydwaddell: a defect was created to take care of this issue

AKobets (approved):

PR #272, opened 2021-07-21
DE44196 Fix due date being cut off

KaiPrince:
## Overview

 Before | After 
------- | ------
 image | image

Fix due date being cut off.

### Reason for change

See Ticket: [DE44196](https://rally1.rallydev.com/#/515253241784d/custom/367300408400?detail=%2Fdefect%2F602631407919&fdp=true?fdp=true): 1. Work To Do Widget Truncates Date for Long Course Names

### Changes summarized

- Remove Flex styling.
- Use `span` instead of `div` because these are now inline elements.
- Wrap all in a `div`.
- Add `text-overflow` property to slot.
- Add `vertical-align` property to bullet.
- Balance spacing on both sides of bullet.

### Suggested future tasks

None.

## Checklist

- [x] No new warnings are introduced
- [x] Comments and documentation are up to date
- [x] TODOs resolved or added to issue tracker
- [x] No commented-out code
- [ ] This isn't a 'quick-and-dirty' job

### DIFF
--- a/features/workToDo/d2l-w2d-attribute-list.js
+++ b/features/workToDo/d2l-w2d-attribute-list.js
@@ -4,10 +4,6 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
 class W2DAttributeList extends SkeletonMixin(LitElement) {
 	static get styles() {
 		return [ super.styles, css`
-			:host {
-				height: 1rem;
-				overflow-y: hidden;
-			}
 			*,
 			::slotted(*) {
 				overflow: hidden;
--- a/features/workToDo/d2l-w2d-attribute-list.js
+++ b/features/workToDo/d2l-w2d-attribute-list.js
@@ -12,6 +12,7 @@ class W2DAttributeList extends SkeletonMixin(LitElement) {
 			::slotted(*) {
 				overflow: hidden;
 				white-space: nowrap;
+				text-overflow: ellipsis;
 			}
 		`];
 	}
--- a/features/workToDo/d2l-w2d-attribute-list.js
+++ b/features/workToDo/d2l-w2d-attribute-list.js
@@ -5,16 +5,11 @@ class W2DAttributeList extends SkeletonMixin(LitElement) {
 	static get styles() {
 		return [ super.styles, css`
 			:host {
-				display: flex;
-				flex-flow: row wrap;
 				height: 1rem;
 				overflow-y: hidden;
 			}
 			*,
 			::slotted(*) {
-				align-items: center;
-				display: flex;
-				flex-flow: row nowrap;
 				overflow: hidden;
 				white-space: nowrap;
 			}
--- a/features/workToDo/d2l-w2d-list-item.js
+++ b/features/workToDo/d2l-w2d-list-item.js
@@ -92,14 +92,14 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 				height:18px;
 				width:18px;
 				vertical-align: top;
-				margin-right: 0.15rem;
+				margin-left: -0.15rem;
 			}
 			.d2l-w2d-list-item-attributes *:first-child::before {
 				content: "";
 				height: 0;
 				width: 0;
 				display: initial;
-				margin-right: 0;
+				margin-left: 0;
 			}
 		`];
 
--- a/features/workToDo/d2l-w2d-list-item.js
+++ b/features/workToDo/d2l-w2d-list-item.js
@@ -149,8 +149,10 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 				<d2l-list-item-content>
 					<d2l-activity-name class="d2l-w2d-list-item-name" href="${this.href}" .token="${this.token}"></d2l-activity-name>
 					<d2l-w2d-attribute-list slot="secondary" class="d2l-w2d-list-item-attributes">
-						${ !this.collapsed ? startDate : null }
-						${this._renderAttributeListCollapsed()}
+						<div>
+							${ !this.collapsed ? startDate : null }
+							${this._renderAttributeListCollapsed()}
+						</div>
 					</d2l-w2d-attribute-list>
 					${ !this.collapsed ? html`<d2l-activity-description slot="supporting-info" href="${this.href}" .token="${this.token}"></d2l-activity-description>` : startDate}
 				</d2l-list-item-content>
@@ -169,10 +171,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 		const type = !this._isCourse ? lithtml`<d2l-activity-type href="${this.href}" .token="${this.token}"></d2l-activity-type>` : null;
 		const courseName = this._isCourse ? lithtml`<span>${this.localize('course')}</span>` : lithtml`<span>${this._parentName}</span>`;
 		return html`
-		<div>
-			${this.collapsed ? dueDate : type}<!--
-			-->${courseName}
-		</div>
+			${this.collapsed ? dueDate : type}${courseName}
 		`;
 	}
 
--- a/features/workToDo/d2l-w2d-list-item.js
+++ b/features/workToDo/d2l-w2d-list-item.js
@@ -170,8 +170,8 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 		const courseName = this._isCourse ? lithtml`<span>${this.localize('course')}</span>` : lithtml`<span>${this._parentName}</span>`;
 		return html`
 		<div>
-			${this.collapsed ? dueDate : type}
-			${courseName}
+			${this.collapsed ? dueDate : type}<!--
+			-->${courseName}
 		</div>
 		`;
 	}
--- a/features/workToDo/d2l-w2d-list-item.js
+++ b/features/workToDo/d2l-w2d-list-item.js
@@ -173,7 +173,8 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 		const type = !this._isCourse ? lithtml`<d2l-activity-type href="${this.href}" .token="${this.token}"></d2l-activity-type>` : null;
 		const courseName = this._isCourse ? lithtml`<span>${this.localize('course')}</span>` : lithtml`<span>${this._parentName}</span>`;
 		return html`
-			${this.collapsed ? dueDate : type}${courseName}
+			${this.collapsed ? dueDate : type}
+			${courseName}
 		`;
 	}
 
--- a/features/workToDo/d2l-w2d-list-item.js
+++ b/features/workToDo/d2l-w2d-list-item.js
@@ -91,6 +91,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 				display: inline-block;
 				height:18px;
 				width:18px;
+				vertical-align: top;
 			}
 			.d2l-w2d-list-item-attributes *:first-child::before {
 				content: "";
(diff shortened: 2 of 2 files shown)


features/workToDo/d2l-w2d-attribute-list.js
```diff
@@ -5,18 +5,14 @@ class W2DAttributeList extends SkeletonMixin(LitElement) {
 	static get styles() {
 		return [ super.styles, css`
 			:host {
-				display: flex;
-				flex-flow: row wrap;
 				height: 1rem;
 				overflow-y: hidden;
```
KaiPrince, line 9: Neither of these are strictly required, and removing them doesn't make any visual difference under normal circumstances. Should they be kept for any reason?

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -162,13 +163,15 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
...
 		let dueDate;
 		if (this._dates.due || this._dates.end) {
-			dueDate = lithtml`<div>${this._dates.due ? this.localize('dueWithDate', 'dueDate', formatDate(this._dates.due, {format: 'shortMonthDay'})) : this.localize('endWithDate', 'endDate', formatDate(this._dates.end, {format: 'shortMonthDay'}))}</div>`;
+			dueDate = lithtml`<span>${this._dates.due ? this.localize('dueWithDate', 'dueDate', formatDate(this._dates.due, {format: 'shortMonthDay'})) : this.localize('endWithDate', 'endDate', formatDate(this._dates.end, {format: 'shortMonthDay'}))}</span>`;
 		}
 		const type = !this._isCourse ? lithtml`<d2l-activity-type href="${this.href}" .token="${this.token}"></d2l-activity-type>` : null;
-		const courseName = this._isCourse ? lithtml`<div>${this.localize('course')}</div>` : lithtml`<div>${this._parentName}</div>`;
+		const courseName = this._isCourse ? lithtml`<span>${this.localize('course')}</span>` : lithtml`<span>${this._parentName}</span>`;
 		return html`
+		<div>
```
KaiPrince, line 171: Alternatively, I could have placed this wrapper in the `attribute-list` file, but I think it belongs here.

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -162,13 +163,15 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 	_renderAttributeListCollapsed() {
 		let dueDate;
 		if (this._dates.due || this._dates.end) {
-			dueDate = lithtml`<div>${this._dates.due ? this.localize('dueWithDate', 'dueDate', formatDate(this._dates.due, {format: 'shortMonthDay'})) : this.localize('endWithDate', 'endDate', formatDate(this._dates.end, {format: 'shortMonthDay'}))}</div>`;
+			dueDate = lithtml`<span>${this._dates.due ? this.localize('dueWithDate', 'dueDate', formatDate(this._dates.due, {format: 'shortMonthDay'})) : this.localize('endWithDate', 'endDate', formatDate(this._dates.end, {format: 'shortMonthDay'}))}</span>`;
```
KaiPrince, line 166: I could have just changed the css `display` property, but I chose to replace `div` with `span` because the property will be correct by default. Plus it just felt right.

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -91,6 +91,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 				display: inline-block;
 				height:18px;
 				width:18px;
+				vertical-align: top;
```
bearfriend, line 94: I'm skeptical that `top` is the best value here, but if it works... 🤷 

Can we just quick run through the possible values in the dev tools to see if any else works, if you haven't already?

features/workToDo/d2l-w2d-attribute-list.js
bearfriend: `overflow-y` can go too?

mbardakovD2L: try it with an activity that has a start date in the future - if this is the component I'm thinking of we might have needed display:flex to get that to display properly. see [slack thread](https://d2l.slack.com/archives/GPYDPVCTC/p1611344940000900)

KaiPrince: Yeah, because it's covered by the `overflow` and `no-wrap` in the list item. I think.

KaiPrince:
REMINDER TO MYSELF: Backport when merged.

dcprime:
Did testing on a local build for this fix, and everything checked out without any problems.

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -162,13 +164,15 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
...
 		}
 		const type = !this._isCourse ? lithtml`<d2l-activity-type href="${this.href}" .token="${this.token}"></d2l-activity-type>` : null;
-		const courseName = this._isCourse ? lithtml`<div>${this.localize('course')}</div>` : lithtml`<div>${this._parentName}</div>`;
+		const courseName = this._isCourse ? lithtml`<span>${this.localize('course')}</span>` : lithtml`<span>${this._parentName}</span>`;
 		return html`
-			${this.collapsed ? dueDate : type}
-			${courseName}
+		<div>
+			${this.collapsed ? dueDate : type}<!--
+			-->${courseName}
```
KaiPrince, line 174: Before | After 
------- | ------
 image | image

There's a tiny extra space that gets added between the `span` tags. Fix is to add a comment.
Details on fix here: https://stackoverflow.com/a/22689952

KaiPrince: It's a little icky, but the go-to fixes wouldn't work: `div` + `inline-block`; `margin:0px`; or `padding:0px`.
I didn't try `div` + `font-size:0` because I like that even less than the little I like this.

features/workToDo/d2l-w2d-list-item.js
KaiPrince: Baseline | Bottom | Middle 
---------|--------|-------
image | image | image

Text-bottom | Text-top | Top
------|------|-----
image | image | image

 Sub | Super 
-------|------
 image | image

KaiPrince: I upped the font-size just to make it easier to see the placement difference. Also to show responsiveness.

features/workToDo/d2l-w2d-attribute-list.js
KaiPrince: @mbardakovD2L Darren checked it out and it was good.

features/workToDo/d2l-w2d-list-item.js
bearfriend: Damn, top does look right.

p.s. This was extremely thorough.

features/workToDo/d2l-w2d-list-item.js
bearfriend: I actually think this makes more sense to be on one line with no whitespace.

features/workToDo/d2l-w2d-attribute-list.js
KaiPrince: Removed them. https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/272/commits/2e35e36bba2d556bf4fd0657480a89e346e64367

features/workToDo/d2l-w2d-list-item.js
mbardakovD2L: I can't tell from just this code how the separator icon/circle/bullets are being rendered, but you want to have the same amount of whitespace before and after them for consistency. (I ran into the same issue [here](https://github.com/BrightspaceHypermediaComponents/activities/pull/1526/files#diff-7a3047c0fe41f95269634c1cb6d4e0d20392b2db92dc9407b8ade41a7ae78c92R193), except I decided to go the other route and add extra whitespace instead of taking it away). Your fix works because the comment gets rid of the whitespace before the bullet, but Danny is right that just putting them on one line with no whitespace has the same effect and looks cleaner

features/workToDo/d2l-w2d-attribute-list.js
dcprime: Verified that activities with start dates in the future are displaying correctly with this fix.

KaiPrince: Collapsed | Full | Full (short name)
-----------|-----|-------------
image | image | image


Actually I think there might be an issue when displayed in full screen. I'll look into it.

KaiPrince: Fix in https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/272/commits/658252c5f576563962a3c37467a3ac748e8d641b

features/workToDo/d2l-w2d-list-item.js
KaiPrince: Made 1 line in https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/272/commits/658252c5f576563962a3c37467a3ac748e8d641b

KaiPrince: image

Now the spacing doesn't match these other components. Working on it..

KaiPrince: **Why is this happening, you ask?**

Well the _Quiz_ text is within a shadow DOM, but the _Course_ text is not. See image:
image

Or it's because `display: block` is adding padding between elements that I can't seem to remove, that `flex` doesn't add.

**What's the solution, you ask?**
Here are the options:

1. Increase the spacing around that last bullet, by adding padding or a shadow dom
2. Find out how to remove the gutters from `block`
3. Go back to flex and try to find a flex-compatible truncation
4. Just...keep it in _\*cringe\*_

bearfriend: Well, wait. What do the designs look like? What is the spacing supposed to be? Seems like the larger spacing is correct, no? Which means just put that `courseText` back on its own line?

mbardakovD2L: the `Quiz` text element seems to extend to the right a little past where the text actually ends, which seems to be where the extra space comes from. I was able to counteract it by adding margins like so:

```
.d2l-w2d-list-item-attributes *::before {
	content: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' width='18px' height='18px' viewBox='0 0 18 18'><circle cx='9' cy='9' r='3' fill='%236E7376' fill-rule='evenodd'/></svg>");
	display: inline-block;
	height:18px;
	width:18px;
	vertical-align: top;
	margin-right: 0.15rem; // my change
}
.d2l-w2d-list-item-attributes *:first-child::before {
	content: "";
	height: 0;
	width: 0;
	display: initial;
	margin-right: 0; // my change
}
```

but the root of the issue I think stems from `d2l-activity-type` being a smidge wider than it needs to be

KaiPrince: It's perfect! It's beautiful!
image

KaiPrince: Just a tiny note, that we are deviating slightly from the original, because the margins became wider in total.

Image from before changes:
image

mbardakovD2L: I haven't tested this but we can probably make that work if we change the 0.15rem margin to a 0 margin, and the 0 margin to a -0.15rem margin (basically subtract -0.15rem from both sides)

it might take some fiddling, 0.15rem was just a random number I came up with that happened to look okay

dcprime:
Just reviewed the latest fix for spacing and it looks good now.

features/workToDo/d2l-w2d-attribute-list.js
KaiPrince: New fix in https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/272/commits/e8c1465dfcb1e9c9a32517d98f91161212e53559

features/workToDo/d2l-w2d-list-item.js
KaiPrince: Fixed in https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/272/commits/db08528dc26cff0d6672e8aeadfd7b00dfb402f0

Change | Result
-------- | --------
image | image

zinabat (approved):

KaiPrince:
> /backport

So yeah that's not a thing here

KaiPrince:
> REMINDER TO MYSELF: Backport when merged.

Done.

PR #277, opened 2021-07-26
DE44542 - Adds the option for activities that have not started yet to be clickable (needed for parent portal)

adamsk34:
# Description
Parent portal needs activities that haven't started yet to still be clickable to I've added an option to do that

# Rally
https://rally1.rallydev.com/#/378960033600d/workviews?Name=%5BBfP%5D%20Activities%20that%20haven't%20started%20yet%20aren't%20clickable&Priority=Normal&Project=https%3A%2F%2Frally1.rallydev.com%2Fslm%2Fwebservice%2Fv2.x%2Fproject%2F378960033600&Severity=Minor%20Problem&detail=%2Fdefect%2F605126742893&iteration=u&rankTo=BOTTOM&view=dd31669c-8a70-406a-8edf-892fc6f74aef&fdp=true?fdp=true

### DIFF
--- a/features/workToDo/d2l-w2d-list.js
+++ b/features/workToDo/d2l-w2d-list.js
@@ -46,13 +46,7 @@ class W2dList extends HypermediaStateMixin(LitElement) {
 		return html`
 			<d2l-list separators="${this.collapsed ? 'none' : 'all'}">
 				${activities.map(activity => html`
-					<d2l-w2d-list-item
-						href="${activity.href}"
-						.token="${this.token}"
-						?collapsed="${this.collapsed}"
-						?allow-unclickable-activities="${this.allowUnclickableActivities}"
-						?clickable-future-activities="${this.clickableFutureActivities}">
-					</d2l-w2d-list-item>
+					<d2l-w2d-list-item href="${activity.href}" .token="${this.token}" ?collapsed="${this.collapsed}" ?allow-unclickable-activities="${this.allowUnclickableActivities}" ?clickable-future-activities="${this.clickableFutureActivities}"></d2l-w2d-list-item>
 				`)}
 			</d2l-list>
 		`;
--- a/features/workToDo/d2l-w2d-list.js
+++ b/features/workToDo/d2l-w2d-list.js
@@ -14,11 +14,12 @@ const rel = Object.freeze({
 class W2dList extends HypermediaStateMixin(LitElement) {
 	static get properties() {
 		return {
-			allowUnclickableActivities: { type: Boolean, attribute: 'allow-unclickable-activities' },
 			category: { type: String },
 			collapsed: { type: Boolean },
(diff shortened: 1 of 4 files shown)


ryantmer (approved):
Yay plumbing through config 😅 Kind of a broader question than this PR, but is there a reason why W2D doesn't generally allow for clickable future activities as it is? I figured if there's an alternate link to the activity, that's what would make it clickable, and if there's an alternate link, it's because you can see it.

features/workToDo/d2l-w2d-list.js
```diff
@@ -45,7 +46,7 @@ class W2dList extends HypermediaStateMixin(LitElement) {
 		return html`
 			<d2l-list separators="${this.collapsed ? 'none' : 'all'}">
 				${activities.map(activity => html`
-					<d2l-w2d-list-item href="${activity.href}" .token="${this.token}" ?collapsed="${this.collapsed}" ?allow-unclickable-activities="${this.allowUnclickableActivities}"></d2l-w2d-list-item>
+					<d2l-w2d-list-item href="${activity.href}" .token="${this.token}" ?collapsed="${this.collapsed}" ?allow-unclickable-activities="${this.allowUnclickableActivities}" ?clickable-future-activities="${this.clickableFutureActivities}"></d2l-w2d-list-item>
```
marcsoiferman, line 49: You split the `d2l-w2d-list` component earlier onto separate lines, should you also do that here for consistency?

jstefaniuk-d2l (approved):

jstefaniuk-d2l:
@ryantmer

My assumption is that in general the alternative links got added to HM, but not gated off based on whether the user has access to see the LMS pages, so the start date thing is a way of blocking off things that the user can't click through to in the LMS

JayArora1997 (approved):

ryantmer:
> @ryantmer
> 
> My assumption is that in general the alternative links got added to HM, but not gated off based on whether the user has access to see the LMS pages, so the start date thing is a way of blocking off things that the user can't click through to in the LMS

Ah fair - links shouldn't be added if they're not accessible to the user, but I can't confidently say that things aren't being added when they shouldn't be. Even more confusing, being able to see things before they are accessible (e.g. Discussions) is a possibility, so determining whether or not users can see things before the start date is... complicated 😬

features/workToDo/d2l-w2d-list.js
adamsk34: That's what I did at first, somehow when I split it onto separate lines it made none of the list items be links. It worked when I put it all on one line and it broke when I put it on multiple even when everything else was exactly the same, and I have no idea why. I decided to stop fighting it and just let it be on one line 🤣

marcsoiferman: Well that's an odd one.

PR #280, opened 2021-07-27
fix: US128538 - Use activity usage link to filter out current quiz in alsoIn

alexli01:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F600219060741&fdp=true?fdp=true


components/activity/list/custom/quiz/d2l-activity-list-item-question.js
```diff
@@ -158,19 +159,11 @@ const componentClass = class extends SkeletonMixin(HypermediaStateMixin(Localize
...
 	_getAlsoInString() {
-		const alsoIn = this._getAlsoIn();
 		let alsoInString = `${this.localize('also_in')} `;
-		for (let i = 0; i < alsoIn.length; i++) {
-			alsoInString = alsoInString + alsoIn[i];
-			if (i < alsoIn.length - 1) {
+		for (let i = 0; i < this.parentCollectionsName.length; i++) {
+			alsoInString = alsoInString + this.parentCollectionsName[i];
+			if (i < this.parentCollectionsName.length - 1) {
 				alsoInString = `${alsoInString}, `;
```
mdgbayly, line 167: This looks good but is this comma separated string concatenation of names locale sensitive?
This stackoverflow post implies list formatting is locale sensitive and has a couple of ideas but nothing that looks particularly usable by us out of the box.

https://stackoverflow.com/questions/39237769/rtl-languages-comma-separated-lists

I don't see anything in the Brightspace-ui/intl library that addresses this either.
https://github.com/BrightspaceUI/intl

Possibly worth a question on #web-platform or you can just assume what you have is good enough.

mdgbayly (approved):

components/activity/list/custom/quiz/d2l-activity-list-item-question.js
mdgbayly: Maybe this std JS lib is well enough supported now for us to use it. Still might be worthwhile asking on #web-platform to see how the Gaudi guys think of this.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat

PR #287, opened 2021-08-03
Add back W2D telemetry

bearfriend:
## Context
Telemetry is an important feature for capturing usage data. The  original work-to-do widget has this feature, so we need to add it back.
- Converts work done on original widget to foundations
- Captures fetch calls and publishes information to telemetry
- Installs the telemetry package

https://rally1.rallydev.com/#/357251704080/detail/defect/604102204993


features/workToDo/d2l-w2d-collections.js
```diff
@@ -380,6 +385,15 @@ class W2dCollections extends LocalizeDynamicMixin(HypermediaStateMixin(LitElemen
 		this.requestUpdate('_pagingTotalResultsOverdue', oldValue);
 	}
 
+	_handleViewAllClick(e) {
+		// sadness
```
zinabat, line 389: lol

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -16,18 +16,8 @@ import { ifDefined } from 'lit-html/directives/if-defined';
...
-	checklist: 'https://checklists.api.brightspace.com/rels/checklist-item',
-	content: 'https://api.brightspace.com/rels/content',
-	date: 'https://api.brightspace.com/rels/date',
-	organization: 'https://api.brightspace.com/rels/organization',
-	organizationHomepage: 'https://api.brightspace.com/rels/organization-homepage',
-	quiz: 'https://api.brightspace.com/rels/quiz',
-	survey: 'https://surveys.api.brightspace.com/rels/survey',
-	topic: 'https://discussions.api.brightspace.com/rels/topic'
-});
+import { rels } from './d2l-w2d-rels.js';
```
zinabat, line 19: Why did you decide to move these? It is normally intentional that these are defined with the component, as it's a quick way to see what the component cares about

features/workToDo/w2dSummonActionObserver.js
```diff
@@ -52,6 +54,14 @@ export class W2dSummonAction extends SirenSummonAction {
...
 				this.routedState.addObservables(observer, route);
 			});
+
+			this.routedState.fetchStatus.waitForNextFetch.then(() => {
+				telemetry.markFetchStart(this._telemetryPage);
+
+				this.routedState._fetchStatus.complete.then(res => {
+					telemetry.markFetchEnd(this._telemetryPage, res.properties.pagingTotalResults);
+				});
+			});
```
zinabat, line 64: Nit: style-wise I think you should use `await` keyword to write the inner promise handler to avoid the nesting, but totally a nit

zinabat (commented):
Looks better than I figured it might

features/workToDo/d2l-w2d-telemetry.js
```diff
@@ -0,0 +1,150 @@
+import Events from 'd2l-telemetry-browser-client';
+import { Rels } from 'siren-sdk/src/hypermedia-constants';
```
zinabat, line 2: I would strongly suggest not importing these constants from `siren-sdk`, as we currently intentionally do not have a dependency on siren-sdk. They should be defined here, in this file, for now.

features/workToDo/d2l-w2d-telemetry.js
```diff
@@ -0,0 +1,150 @@
+import Events from 'd2l-telemetry-browser-client';
```
zinabat, line 1: You'll need to install this package

features/workToDo/d2l-w2d-telemetry.js
```diff
@@ -0,0 +1,150 @@
...
+import { Rels } from 'siren-sdk/src/hypermedia-constants';
+
+const W2D_TELEMETRY_ID = 'worktodo';
```
zinabat, line 4: please convert these consts to not use snake case or capitalization - consts are usually defined with `Object.freeze` in this repo, so I will suggest the same here

zinabat:
Please also add links in the description for where this code originated from and the why behind it. We know the context now, but when we read this in a year I would like to have some breadcrumbs

features/workToDo/d2l-w2d-collections.js
```diff
@@ -327,7 +328,7 @@ class W2dCollections extends LocalizeDynamicMixin(HypermediaStateMixin(LitElemen
 
 		return html`
 			${this._renderSkeleton()}
-			${(!overdue || overdue.length === 0) && (!categories || categories.length === 0) ? emptyList : lists}
+			${(overdue && overdue.length) || (categories && categories.length) ? lists : emptyList}
```
mbardakovD2L, line 331: will probably cause conflicts with my fix for the same thing, but you have my blessing to ignore my code and just use your own

bearfriend: Should be rebased and happy as a clam.

features/workToDo/d2l-w2d-list-item.js
bearfriend: Initially, I think I was trying to use them to map back to what request was happening, but then I didn't end up doing that and used the page stuff. I can certainly revert this part.

zinabat: yeah no problem if you have a plan to use the rels elsewhere, just want to make sure we're consistent :D

features/workToDo/d2l-w2d-telemetry.js
```diff
@@ -0,0 +1,156 @@
...
+		this._marks = {};
+		this._createClient();
+	}
+
+	async logActivityNavigatedTo(href, type) {
+		await this._logTelemetryEvent('NavigatedTo', href, type);
+	}
+
+	async logViewAllClicked(href) {
+		await (await this._logTelemetryEvent('ViewAllClicked', href, 'ViewAll'));
```
bearfriend, line 41: wtf have I done?

`return`? Is it creating a default promise? _What is you doin'?_

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -361,13 +380,15 @@ class W2DListItemSurvey extends W2DListItemMixin {
 				type: String,
 				observable: observableTypes.link,
 				rel: 'alternate',
-				route: [{observable: observableTypes.link, rel: rels.survey}],
+				route: [{observable: observableTypes.link, rel: rels[this.activityType]}],
```
zinabat, line 383: didn't we figure out that you can't do this unless you use `this.constructor.activityType`? Accessing a static property this way is odd

bearfriend: Because `properties` is also `static`, `this` is not an instance but the class itself here. In `_handleItemLinkClick` I access it from the constructor because in that case `this` is the instance.

bearfriend: Screen Shot 2021-08-04 at 11 11 06 AM

zinabat: pff, javascript you drunken fart of a language.
makes sense, thanks 😂

package.json
```diff
@@ -47,6 +47,7 @@
     "@brightspace-ui/intl": "^3.5.0",
     "d2l-course-image": "github:Brightspace/course-image#semver:^3",
     "d2l-navigation": "github:BrightspaceUI/navigation#semver:^4",
+    "d2l-telemetry-browser-client": "github:Brightspace/d2l-telemetry-browser-client",
```
zinabat, line 50: Should this be versioned?

mbardakovD2L (approved):

zinabat (approved):

PR #294, opened 2021-08-09
fix: Get resolved token for simple components

BhamelinD2L:
Small fixes for enrollment rules UI: 

- Gets the resolved token and passes it to the non-hmc component `d2l-discover-rule-card` so it has a proper token
- Changes the enrollment rules commit to use the `id` instead of the `type` during rule generation
- Updates tests


features/discover/d2l-discover-rules.js
```diff
@@ -21,6 +21,7 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
 			isSelfEnrollable: { type: Boolean, observable: observableTypes.classes,
 				method: (classes) => classes.includes(rels.selfAssignableClass),
 				route: [{observable: observableTypes.link, rel: rels.organization }] },
+			_resolvedToken: {type: String },
```
bearfriend, line 24:
```suggestion
			_resolvedToken: { type: String },
```

bearfriend (approved):
:shipit:

PR #299, opened 2021-08-18
fix: Bhamelin/de44731 checkbox

BhamelinD2L:
The discover checkbox component now checks the `self-assignable` class from the BFF's course route to determine whether to check the checkbox or not.

The `organizations` route that I've removed references to in this PR could not be used for `self-assignable` because its header says to cache the response within the browser for 300 seconds. Since we are using a save->refresh page->get value workflow, that caching causes some issues.

We don't need that route for anything else, so I've deleted it from tests and demos.

Tests were slightly edited, because this small change somehow broke them all. Waiting until rules are initialized before spying on the commit takes care of the issues.

Tested in ArtemisDev via BSI.


features/discover/d2l-discover-rules.js
```diff
@@ -10,17 +10,17 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
...
 const profileCount = 3;
 
 class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStateMixin(LitElement))) {
 	static get properties() {
 		return {
-			isSelfEnrollable: { type: Boolean, observable: observableTypes.classes,
-				method: (classes) => classes.includes(rels.selfAssignableClass),
-				route: [{observable: observableTypes.link, rel: rels.organization }] },
+			isSelfEnrollable: {
+				type: Boolean, observable: observableTypes.classes,
```
dkong-d2l, line 21: nit: why doesn't `observable` deserve its own line?

dkong-d2l (approved):

PR #300, opened 2021-08-24
Fix w2d serge/localization

bearfriend:
I'm fairly confident in this but not 100%, so I'm limiting this PR to just Work To Do. If it works we'll fix the rest of the repo.


foundation-components.serge.json
```diff
@@ -294,18 +294,17 @@
     },
     "output_lang_rewrite": [
       "ar-sa ar",
+      "cy-gb cy",
+      "da-dk da",
       "de-de de",
       "es-mx es",
```
mbardakovD2L, line 300: probably not super relevant, but why does `es-mx` get remapped to `es`, whereas `es-es` is its own thing?

bearfriend: Back before any of us were born all of D2L's Spanish-speaking clients were in the Americas and there was no `es-es` langpack, so it seemed logical to set `es-mx` as the base `es` file. Same goes for `fr-ca` -> `fr`. It's a mess now, and something I'd ultimately like to change, but for now this is what we have 😢 .

zinabat: fun

zinabat (approved):

PR #312, opened 2021-09-17
fix: remove #empty in worktodo activities rel to make use of cache 

klhuff:
With this change, the intitial request to the LMS will return data. But more importantly, the activities will be cached, so that the second requests to both activities and overdue api don't need to both calculate all activities.

This roughly halves the work that the worktodo widget does on the LMS end.

### DIFF
--- a/features/workToDo/d2l-w2d-work-to-do.js
+++ b/features/workToDo/d2l-w2d-work-to-do.js
@@ -10,7 +10,7 @@ import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynam
 import { telemetry } from './d2l-w2d-telemetry';
 
 const rel = Object.freeze({
-	myActivities: 'https://activities.api.brightspace.com/rels/my-activities#empty',
+	myActivities: 'https://activities.api.brightspace.com/rels/my-activities',
 	myOrganizationActivities: 'https://activities.api.brightspace.com/rels/my-organization-activities#empty',
 	organization: 'https://api.brightspace.com/rels/organization',
 	organizationHomepage: 'https://api.brightspace.com/rels/organization-homepage',
--- a/test/data/workToDo.js
+++ b/test/data/workToDo.js
@@ -10,7 +10,7 @@ export const workToDoMain = {
 	links: [
 		{
 			rel: [
-				'https://activities.api.brightspace.com/rels/my-activities#empty',
+				'https://activities.api.brightspace.com/rels/my-activities',
 			],
 			href: '/w2d-activities',
 		},


features/workToDo/d2l-w2d-work-to-do.js
```diff
@@ -10,7 +10,7 @@ import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynam
 import { telemetry } from './d2l-w2d-telemetry';
 
 const rel = Object.freeze({
-	myActivities: 'https://activities.api.brightspace.com/rels/my-activities#empty',
+	myActivities: 'https://activities.api.brightspace.com/rels/my-activities',
```
klhuff, line 13: This means that the initial requests to activities api will do work, where previously it was just returning the links for activities and overdue.

However, because this will add all activities for the user to the cache, the subsequent api calls do next to no work.

Previously, because both these api calls occur simultaneously (the activities and overdue api), neither could make use of the caching, and thus we calculated all activities twice.

lloydwaddell:
> This roughly halves the work that the worktodo widget does on the LMS end.

Are there API changes to go with this? My understanding is that the empty call is only used to generate the hypermedia response (links / actions) necessary for w2d to load. There isn't any DB interactions.

klhuff:
> > This roughly halves the work that the worktodo widget does on the LMS end.
> 
> Are there API changes to go with this? My understanding is that the empty call is only used to generate the hypermedia response (links / actions) necessary for w2d to load. There isn't any DB interactions.
> 
> https://search.d2l.dev/xref/lms/le/activities/D2L.LE.Activities.Extensibility/Api/Hypermedia/ActivitiesController.cs?r=cc2cdcc1#481

You're correct. But the actual problem is that the widget then takes both links that it gets back (activities, and overdue) and calls them both simultaneously. Because they're called simultaneously, there is no caching, and thus the LMS builds the domain objects for every single activity twice (one in each of two different api calls).

This changes makes us do the work of building all of the activities upfront in the initial call to get those links. But, more importantly, puts them into the cache before the two api calls are made. And with that, we only build all of the activities once.

features/workToDo/d2l-w2d-work-to-do.js
jopnick: The alternative to this would be to create a _new_ `rel` which we statically format for _all_ the requests, `https://activities.api.brightspace.com/rels/activities#two-weeks` which basically pre-formats the action with the dates we want and we just attach it as a link.

you get a server-formated date range that matches what you need and we just use the first response

jopnick: This _would_ involve a server-side change and the corresponding client side change, but it would remove the throwing away of the data, get you the correct data-set with the right range, as well as prime the cache

klhuff: I'll admit I'm pretty indifferent to the thrown away data bit. My main concern is reducing the number of times we build the domain objects for all activities and thus reduce the number of queries hammering our SQL servers.

lloydwaddell: I am not opposed to this as a temp solution, but If we are going to be loading the data up-front and then throwing it away. I would be inclined to say we should create a new API to merge upcoming/overdue calls into one api. The only difference is dropping the forced IFilters. We would only then filter the data once and return half the items.

jopnick: why would this need a whole new API? I feel like the cost of adding a new rel, with the hot-path the data you need, so long as the server generates this link and not the client, then this is veyr low-overhead, very low effort and does what we need with no waste

lloydwaddell: I'm just thinking it could take a step further, the #two-week rel will only return activities from now + two weeks, which is data that is unnecessary if you have 20 overdue activities. We are calling 2 APIs that basically do the same work, give me all the activities and then limits based on upcoming or overdue, so why not do it once for all activities.

klhuff:
Still looking for approval on this one if possible.

lloydwaddell (approved):
as long as @jopnick has no strong feelings on the rel not being #two-weeks

zinabat:
🚢 it

jopnick (approved):
I think just this client-side change is enough to fix the cache-miss, even if it throws away the results of the initial query. I think the conversation about what server-side changes, to make this more efficient can continue w/o blocking this first step :shipit:

PR #313, opened 2021-09-17
After adding another condition focus is on new one

ChrisLabattD2L:
[DE45110](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fdefect%2F608843206171)

After clicking on the "Add another condition", the focus is now on the latest condition rather than remaining on the button previously clicked to improve the accessibility.

### DIFF
--- a/features/discover/d2l-discover-rule-picker.js
+++ b/features/discover/d2l-discover-rule-picker.js
@@ -140,12 +140,6 @@ class RulePicker extends MatchCountMixin(LocalizeDynamicMixin(HypermediaStateMix
 			this._updateMatchCount();
 			this._sizeChanged();
 		}
-
-		const conditionOrgUnitIDList = this.shadowRoot.querySelectorAll('select');
-		if (conditionOrgUnitIDList.length > 1) {
-			const lastOrgUnitID = conditionOrgUnitIDList[conditionOrgUnitIDList.length - 1];
-			lastOrgUnitID.focus();
-		}
 	}
 
 	reload(newConditions) {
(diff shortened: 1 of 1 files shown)


BhamelinD2L (changes requested):
Because this is in updated, it will run every single time there is more than 1 condition and any aspect of the component changes. This means it will constantly focus the last condition any time you interact with any of the the attributes in the picker.

You should put the code changes under the add condition button handler, await updateComplete, and then focus.

ChrisLabattD2L:
> Because this is in updated(), it will run every single time there is more than 1 condition and any aspect of the component changes. This means it will constantly focus the last condition any time you interact with any of the the attributes in the picker.
> 
> You should put the code changes under the add condition button handler, await updateComplete, and then focus.

The changes I just pushed should function properly, let me know if there are any other changes you think are necessary 😄

features/discover/d2l-discover-rule-picker.js
```diff
@@ -164,8 +164,15 @@ class RulePicker extends MatchCountMixin(LocalizeDynamicMixin(HypermediaStateMix
 		this._sizeChanged();
 	}
 
-	_addDefaultCondition() {
+	async _addDefaultCondition() {
 		this._addCondition();
+
+		await this.updateComplete;
+		const conditionOrgUnitIDList = this.shadowRoot.querySelectorAll('select');
+		if (conditionOrgUnitIDList.length > 1) {
```
BhamelinD2L, line 172: Is there a reason we need this check? What happens if we remove it?

ChrisLabattD2L: I included the check to reduce the risk of an error being caused due to the query selector not being able to find anything

ChrisLabattD2L: Tested it after removing it and it works the same, it was mainly put in as a precaution because it was previously in updated(), so it may not be necessary anymore.

BhamelinD2L (approved):

PR #319, opened 2021-09-27
feat: Added limits to rules, conditions and attributes

ChrisLabattD2L:
Part of [US126159](https://rally1.rallydev.com/#/357252275780d/dashboard?detail=%2Fuserstory%2F512835224632), which adds limits to the rules, condition and attributes.

The changes for the attribute-picker are in this PR: https://github.com/BrightspaceUILabs/attribute-picker/pull/55


features/discover/d2l-discover-rule-picker.js
```diff
@@ -35,7 +36,10 @@ class RulePicker extends MatchCountMixin(LocalizeDynamicMixin(HypermediaStateMix
 			rules: { type: Array },
 			_defaultType: { type: String },
 			_matchCount: { type: Number },
-			_getMatchCount: { observable: observableTypes.summonAction, name: 'match-count' }
+			_getMatchCount: { observable: observableTypes.summonAction, name: 'match-count' },
+			_conditionsFull: { type: Boolean },
+			attributeLimit: { type: Number, attribute: 'attribute-limit' },
```
BhamelinD2L, line 41: We don't need to take the attribute limit or condition limit all the way to the top component. You've defaulted it to 10, so we can keep it simple and use that default

features/discover/d2l-discover-rule-picker.js
```diff
@@ -172,6 +184,10 @@ class RulePicker extends MatchCountMixin(LocalizeDynamicMixin(HypermediaStateMix
 		const conditionOrgUnitIDList = this.shadowRoot.querySelectorAll('select');
 		const lastOrgUnitID = conditionOrgUnitIDList[conditionOrgUnitIDList.length - 1];
 		lastOrgUnitID.focus();
+
+		if (this.conditions.length >= this.conditionLimit) {
+			this._conditionsFull = true;
```
BhamelinD2L, line 189: We don't need this as a property - Your check can be its own function and can be used in place of the property.

features/discover/d2l-discover-rule-picker.js
```diff
@@ -243,6 +259,10 @@ class RulePicker extends MatchCountMixin(LocalizeDynamicMixin(HypermediaStateMix
 			condition.properties.state = conditionStates.remove;
 			this._updateMatchCount();
 		}
+
+		if (this.conditions.length <= this.conditionLimit) { // It's <= 10 because the condition hasn't been removed yet, check could potentially be removed if agreed upon
```
BhamelinD2L, line 263: This is why it's more convenient to just do this check when we actually need it instead of working around it being another property - We already have everything we need to determine it when needed and without this additional code

features/discover/d2l-discover-rules.js
```diff
@@ -30,7 +31,11 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
 			_dialogOpened: { type: Boolean },
 			_entitlementsHref: { observable: observableTypes.link, rel: rels.entitlementRules },
 			_createEntitlement: { observable: observableTypes.action, name: 'create-entitlement-rules' },
-			_getEntitlement: { observable: observableTypes.summonAction, name: 'entitlement-rules' }
+			_getEntitlement: { observable: observableTypes.summonAction, name: 'entitlement-rules' },
+			_rulesFull: { type: Boolean },
+			ruleLimit: { type: Number },
+			attributeLimit: { type: Number },
+			conditionLimit: { type: Number }
```
BhamelinD2L, line 38: Same deal as previous comments - We probably don't need the boolean, and the limit counters should just take advantage of the default. 

We aren't going to be changing the limit count or reusing this discover-rules specific component elsewhere, so I don't think there is value in the added complexity of carrying these properties forward through each level.

ChrisLabattD2L:
I've changed the properties into functions like you recommended, but I think it might be worthwhile talking about keeping the propagating properties for the limits in case we want to let users choose what the limit is. Otherwise, I agree that having them there just adds more complexity than needed

BhamelinD2L (changes requested):
This all looks good!
As the last part of this, you should update the tests to ensure this behavior is retained in case of future changes. Your tests should ensure that assigning the limits to some non-10 value has the intended effect, and that you can't add n + 1 items  - Buttons are disabled, alerts are shown, etc.

PR #329, opened 2021-10-18
feat: d2l-discover-date-picker

ChrisLabattD2L:
Based on [US132158](https://rally1.rallydev.com/#/611579333303d/dashboard?detail=%2Fuserstory%2F610209547387)

The integration of the d2l-discover-date-picker component into the rule picker dialog will be a different PR

This PR will have to wait to be merged once [this PR](https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/326) gets merged into master.


features/discover/test/d2l-discover-date-picker.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+	describe('accessibility', () => {
+		it('should pass all aXe tests', async() => {
+			const el = await fixture(html`<d2l-discover-date-picker></d2l-discover-date-picker>`);
+			expect(el).to.be.accessible();
+		});
+	});
+
+	describe('rendering', () => {
+		it('initialize default component', async() => {
```
BhamelinD2L, line 28: nit: The text should read as a proper sentence. 'it initialize**s** the default component, 'it change**s** between single and range dates', etc

BhamelinD2L (changes requested):
You should add a demo for the component so that it can be reviewed and tested in isolation so we can test and catch any problems now before we add the complexity of integration. Looks good so far though!

features/discover/d2l-discover-date-picker.js
```diff
@@ -0,0 +1,142 @@
...
+				@change="${this._onDatePickerChanged}"
+				id="picker-date">
+				<option value="before" ?selected="${this.dateType === dateTypeEnum.before}">${this.localize('text-date-picker-before')}</option>
+				<option value="range" ?selected="${this.dateType === dateTypeEnum.range}">${this.localize('text-date-picker-range')}</option>
+				<option value="after" ?selected="${this.dateType === dateTypeEnum.after}">${this.localize('text-date-picker-after')}</option>
+			</select>
+
+			<div class="d2l-date-field">
+				${this._datePickerChecker() ? html`
+					<d2l-input-date-range label="Date Range" child-labels-hidden label-hidden required start-value="${this.primaryDate}" end-value="${this.secondaryDate}" @change="${this._onRangeDateChange}"></d2l-input-date-range>
```
BhamelinD2L, line 87: If just a line or two is huge like this, it's more readable to newline each attribute so that it doesn't need to be scrolled over to read

features/discover/d2l-discover-date-picker.js
```diff
@@ -0,0 +1,142 @@
...
+
+		this.dispatchEvent(new CustomEvent('d2l-date-changed', {
+			bubbles: true,
+			composed: true,
+			detail: {
+				datePicker: [this.dateType, this.primaryDate, this.secondaryDate]
+			}
+		}));
+	}
+}
```
BhamelinD2L, line 140: This should be happening when you change the dropdown off of range. You're removing one of the dates, but the data doesn't reflect it has been removed. 

Currently if you set two dates, switch `range` to `before` and save, the last fired event will be the range and the user will not save what the UI reflects.

features/discover/test/d2l-discover-date-picker.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+		});
+
+		it('sends correct data after changing from a ranged date to single date', async() => {
+			const el = await fixture(html`<d2l-discover-date-picker date-type="range" primary-date="2020-12-12" secondary-date="2021-01-01"></d2l-discover-date-picker>`);
+			expect(el).to.be.not.null;
+			el.dateType = dateTypeEnum.after;
+
+			const listener = oneEvent(el, 'd2l-date-changed');
+			el._updateRulePicker();
```
BhamelinD2L, line 88: Calling this function cheats the event and prevents you from finding the above noted bug. It doesn't adequately represent a real workflow.
In general try to avoid calling private functions in tests. It's much better to simulate user actions like modifying the select and awaiting changes whenever possible.

BhamelinD2L (changes requested):

features/discover/d2l-discover-date-picker.js
```diff
@@ -0,0 +1,142 @@
...
+			</div>
+		`;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('_getPossibleValues')) {
+			this._updatePossibleValues();
+		}
+	}
```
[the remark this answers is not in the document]
ChrisLabattD2L: Thanks for pointing this out, I totally forgot to remove it!

BhamelinD2L (approved):
Looks great now. Thank you!

PR #334, opened 2021-10-25
feat: Us126153 polish animate

ChrisLabattD2L:
Part of [US126153](https://rally1.rallydev.com/#/611579333303d/custom/367300408400?detail=%2Fuserstory%2F512827833300)

Adds the animate directive to the discover rule cards when they get added or removed.


features/discover/d2l-discover-rule-picker-dialog.js
```diff
@@ -117,23 +124,29 @@ class RulePickerDialog extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixi
...
-				properties: {
-					matchCount: picker._matchCount,
-					userList: picker._userList,
+				rule:{
+					entities: [...picker.conditions],
+					properties: {
+						matchCount: picker._matchCount,
+						userList: picker._userList,
+					},
+					rel: [rels.rule]
```
BhamelinD2L, line 133: Restructuring the rule solely to store an animation state isn't ideal. We also want the object structure to make clear sense - Why does a rule store a rule now? What does that mean? Which one do I actually want?

Can we just store the new `state` property within the rule itself, rather than making it in a parent that stores the rule as well?

ChrisLabattD2L: I'll be honest, I don't remember why I decided to do this... I think it was because I was trying to recreate the sample animate directive. I've changed the way it stores state so that it resembles what it looked like previously. 😄

BhamelinD2L (approved):

PR #340, opened 2021-11-03
fix: use daylight savings time safe version of adding days

klhuff:
Previous version of this was taking a number of days and multiplying by the number of milliseconds in a day. This is a problem because the days around daylight saving's time are either 25 or 23 hours long.


grant-cleary (approved):

grant-cleary:
> /backport

Unless this repo did its own thing, I doubt it has the `backport` action. I imagine you'll need to backport it manually.

KaiPrince:
Wow yes of course it won't I see my error now

PR #343, opened 2021-11-05
Add messageformat-validator and fix remaining issues

bearfriend:
Added `messageformat-validator` language file linting to the `lint` script and fixed any issues that it reported.


package.json
```diff
@@ -7,9 +7,13 @@
   },
   "version": "0.59.2",
   "scripts": {
-    "lint": "npm run lint:eslint && npm run lint:lit",
+    "lint": "npm run lint:eslint && npm run lint:lit && npm run lint:langs",
```
mbardakovD2L, line 10: nit: we called this `locales` in manager-view-fra, do we want that to be consistent across repos?

mbardakovD2L (approved):
This looks pretty consistent with how it's used in manager-view-fra, barring a couple differences I looked up via `npx mfv --help` that all look reasonable. I think this is good to go

package.json
bearfriend: Yeah I think they should all be `langs` since that's the pattern for the directories in most repos. In MD they are in `locales/` but we can change that.

package.json
```diff
@@ -11,9 +11,8 @@
     "lint:eslint": "eslint . --ext .js,.html",
     "lint:langs": "mfv -e -i untranslated",
     "lint:lit": "lit-analyzer  \"{components,demo,features,test}/**/*.js\" --strict --rules.no-unknown-tag-name off --rules.no-incompatible-type-binding off --rules.no-unknown-attribute off --rules.no-unknown-property off",
-    "locales:add": "mfv add-missing",
-    "locales:rename": "mfv rename",
-    "locales:remove": "mfv remove-extraneous -s en",
```
bearfriend: The `-s en` wasn't necessary after I added `mfv.config.json` with a `source` field.

PR #344, opened 2021-11-05
US132870  Handle New Discover possible-values Entity

dkong-d2l:
The Discover BFF intends to return a condition values as `named-value` entities so that resource IDs ((such as user IDs, or role IDs) are not shown to the user.  While this PR does not do that, it will accept these new value entities and otherwise function just as it did before.  This will be used as a stepping stone to fully converting the UI to not showing resource IDs to the user while still passing said IDs back to the BFF.  It will, at least, allow the [Discover BFF PR](https://github.com/Brightspace/discovery-bff/pull/159) to be merged...


dkong-d2l:
Curiously, changing `type` to `name` [here](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/discover/test/d2l-discover-attribute-picker.test.js#L39) didn't fail any tests.  Shouldn't it have?

BhamelinD2L:
> Curiously, changing `type` to `name` [here](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/discover/test/d2l-discover-attribute-picker.test.js#L39) didn't fail any tests. Shouldn't it have?

There isn't a test for whether the name is displayed or not, but there should be given how simple it is. I can add some in with my lang PR very soon.

That being said, like this in UX are often done by reviewing the demo, because tests alone really can't confirm the text isn't placed incorrectly, or cropped, or written yet completely invisible because of css, etc. You really do need to see it.

For this reason, take a look at these demo response files, [One](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/discover/demo/condition-typeb-values.json) [Two](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/discover/demo/condition-typea-values.json#L3) [Three](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/discover/demo/course-entitlement-rules-2.json) and modify them to suit your new formatting - If the demos from `npm run start` still show your changed data, we should be in good shape.

features/discover/test/d2l-discover-attribute-picker.test.js
```diff
@@ -66,21 +67,15 @@ const conditionTypeC = {
...
-		values: assignableValuesB
-	}
+	entities: createValueEntities(assignableValuesB)
 };
 
 const possibleValuesC = {
-	properties: {
-		values: assignableValuesC
-	}
+	entities: createValueEntities(assignableValuesC)
```
BhamelinD2L, line 78: I can't leave a comment at the right place in the code, but replace the last test in thisn file (line 198) with the following edit and your tests should pass. It seems your changes delayed things very slightly for safari, and it started being flaky. This ensures the inner picker is fully ready before it's sent the new attribute.
```
		it('fires the d2l-attributes-changed event when the inner attribute list is modified', async() => {
			//Wait for the assignable attributes update to go through
			await waitUntil(() => el.assignableAttributes && el.assignableAttributes.length > 0, 'assignableAttributes not initialized');
			const innerPicker = el.shadowRoot.querySelector('d2l-labs-attribute-picker');
			await innerPicker.updateComplete;

			const  listener = oneEvent(el, 'd2l-attributes-changed');
			el.addAttribute('Zebra');
			const result = await verifyEventTimeout(listener, 'no event fired');
			expect(result).to.not.equal('no event fired');
			expect(result.detail.attributeList[2]).to.equal('Zebra');
		});
```

BhamelinD2L (approved):
Works correctly with both versions of the data as promised and looks good to me!

PR #345, opened 2021-11-09
Us132680 localize default attributes

BhamelinD2L:
[US132680](https://rally1.rallydev.com/#/?detail=/userstory/611074386361&fdp=true): Discover UI - Add localization for default user attributes

- Creates a new mixin, `DiscoverLocalizationMixin` that extends the basic `LocalizeDynamicMixin` and adds a simple function that reviews attribute type IDs and values, and localizes them if they are part of the core list that requires localization. 
- Currently this list is just _orgRole.
- Added tests to ensure it is applied to the card titles as well as the dialog attribute lists.


features/discover/lang/ar.js
```diff
@@ -8,6 +8,7 @@ export default {
 	"label-checkbox": "توفير هذا المقرر التعليمي في أداة الاكتشاف ليتمكن المتعلّمون من إتمام عملية التسجيل الذاتي", // label for the discover checkbox
 	"label-condition-type": "نوع الشرط", // label for screen readers for the condition type select box
 	"label-condition-value": "قيمة الشرط", // label for screen readers for the condition value input
+	"select-org-role": "Role in Organization", // label for selecting a collection of users by their roles in the LMS
```
ChrisLabattD2L, line 11: This text localization already exists under the name `text-role-in-organization` (line 23) so you don't have to add it again. (same goes for the rest of the languages)

features/discover/test/d2l-discover-rule-card.test.js
```diff
@@ -9,8 +9,8 @@ const rels = Object.freeze({
 
 const rule =  {
 	entities: [
-		{ properties: { type: 'Fruit', values: ['apple', 'orange'] }, rel: [rels.condition] },
-		{ properties: { type: 'Entree', values: ['spaghetti'] }, rel: [rels.condition] }
+		{ properties: { type: 'Fruit', id: '_fruit', values: ['apple', 'orange'] }, rel: [rels.condition] },
+		{ properties: { type: 'Entree', id: '_orgRole', values: ['55'] }, rel: [rels.condition] }
```
ChrisLabattD2L, line 13: Small nitpick, but maybe switch the type from `Entree` to something like `Organization Role` so that the type is consistent with the ID

features/discover/test/d2l-discover-rule-picker.test.js
```diff
@@ -35,7 +35,20 @@ const entitlementEntity = {
 const conditionTypesEntity = {
 	entities: [
 		{ rel: [rels.conditionType], properties: { type: 'Fruit', id: '_fruit' } },
-		{ rel: [rels.conditionType], properties: { type: 'Entree', id: '_entree' } }
+		{ rel: [rels.conditionType], properties: { type: 'Entree', id: '_entree' } },
+		{ rel: [rels.conditionType], properties: { type: 'Should not be displayed', id: '_orgRole' } }
```
ChrisLabattD2L, line 39: I just want to double check the purpose of this last conditionType is, and why the type is `Should not be displayed`, is it because the conditionType name won't matter as long as you have an id that is found in the `localizeableTypes` dictionary?

features/discover/mixins/discover-localization-mixin.js
```diff
@@ -0,0 +1,20 @@
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
+
+const localizeableTypes = Object.freeze({
```
ChrisLabattD2L, line 3: Another small nitpick, but `localizeable` should be spelt `localizable`

features/discover/mixins/discover-localization-mixin.js
```diff
@@ -0,0 +1,20 @@
...
+});
+
+export const DiscoverLocalizationMixin = superclass => class extends LocalizeDynamicMixin(superclass) {
+	static get localizeConfig() {
+		return {
+			importFunc: async lang => (await import(`../lang/${lang}.js`)).default
+		};
+	}
+
+	_getLocalizedAttributeText(conditionId, conditionName) {
```
ChrisLabattD2L, line 14: In the localize mixin from core, the function `localize` is public (no underscore), should this function also be public?

ChrisLabattD2L (changes requested):
I noticed that the value of the `select-org-role` localization already exists as the `text-role-in-organization` localization value, so you don't need to add a whole new one.

Other than that it's mostly just minor nitpicks + clarification on something.

features/discover/lang/ar.js
BhamelinD2L: I rebased and I didn't notice, thanks

features/discover/test/d2l-discover-rule-picker.test.js
BhamelinD2L: Yeah. the value of type is irrelevant because we only care about the ID for core attributes, and it's more clear with a name like this in the test data.

features/discover/test/d2l-discover-rule-card.test.js
BhamelinD2L: I have made it match the below version, `should not be displayed`.

features/discover/mixins/discover-localization-mixin.js
BhamelinD2L: It should and now is, thanks.

ChrisLabattD2L (commented):
Looks good now, you'll just have to resolve the conflicts, then you'll be good to merge!

ChrisLabattD2L (approved):

PR #346, opened 2021-11-10
US125522 Switch attribute picker components

BhamelinD2L:
[US125522](https://rally1.rallydev.com/#/?detail=/userstory/505571679024&fdp=true): Discover UI - Migrate Attribute Picker component in to the Multi-select component Switches attribute pickers to use the new multi select version of the picker.
All existing tests appear to pass, and working through the demos things appears to be behaving as you'd expect.


ChrisLabattD2L (approved):
Multi-select Tool: Upgraded
Long have we waited
Attribute Picker: Migrated

PR #347, opened 2021-11-10
Us131416 coi skellies

remytruong:


features/discover/d2l-discover-rule-card.js
```diff
@@ -78,7 +84,7 @@ class RuleCard extends DiscoverLocalizationMixin(RtlMixin(LitElement)) {
 			<d2l-card class="d2l-rule-card" .animate="${animateAction}">
 				<div slot="content">
 					<div>
-						<div class="d2l-rule-card-title">${this._title}</div>
+						<div class="d2l-rule-card-title d2l-skeletize">${this._title}</div>
 						<d2l-discover-rule-card-match-info token="${this.token}" .rule="${this.rule}"></d2l-discover-rule-card-match-info>
```
ChrisLabattD2L, line 88: It may be worthwhile to also skeletize the content within `d2l-discover-rule-card-match-info` and try to skeletize within there rather than from directly in `d2l-profile-image` since we're not the only ones to use that component, and it could cause issues in other uses. Also, be sure to skeletize the text for the match count, and have it be a separate skeleton if possible.

It would also be a good idea to skeletize `d2l-discover-rule-list` since that component functions very similarly to this one, it just has read-only aspects to it.

features/discover/d2l-discover-rule-card.js
```diff
@@ -78,7 +84,7 @@ class RuleCard extends DiscoverLocalizationMixin(RtlMixin(LitElement)) {
 			<d2l-card class="d2l-rule-card" .animate="${animateAction}">
 				<div slot="content">
 					<div>
-						<div class="d2l-rule-card-title">${this._title}</div>
+						<div class="d2l-rule-card-title d2l-skeletize">${this._title}</div>
```
ChrisLabattD2L, line 87: You may need to alter the css of this skeleton since it overflows onto the `d2l-dropdown-context-menu` component (hover over the downwards facing arrow to see what I mean). Though it may also be worthwhile to hide the arrow until the content has loaded in, maybe try to find out if `d2l-dropdown-context-menu` already has a pre-made skeleton that you can use?

features/discover/d2l-discover-rules.js
```diff
@@ -267,6 +287,17 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
...
+	_renderSkeletonRules() {
+		return html`
+			<div class="d2l-discover-rule-card-skeleton">
+			</div>
+			<div class="d2l-discover-rule-card-skeleton">
+			</div>
+			<div class="d2l-discover-rule-card-skeleton">
+			</div>
+		`;
+	}
```
ChrisLabattD2L, line 299: I'm not sure if always loading in 3 skeletons is always a good idea, may be best to only load in one since it would be weird to have 3 skeletons if the user only has 1 or 2 rules set up.

ChrisLabattD2L (changes requested):
It's a good start, but there are still some components that need to be skeletized, also it's a good idea to only make changes within this repo if possible, so avoid making changes within the `d2l-users` repo unless you really have to since other teams use it as well and our changes could negatively affect them. Making changes to the `checkbox-drawer` repo is a different situation since we're the only team using it at the moment, but I've already left my comments there for that aspect.

I didn't look too hard into the `d2l-discover-rules` file yet, as I felt that focusing more on the sub-components may be a good place to start. However I did notice that the skeletons for the rules look like this: 

image

You might want to work on adding more details to it so that it looks more like what the cards look like once they get loaded in, and not just be big squares. (Also on a side not, they don't have the skeleton pulsing effect for me, but I'm not too sure if that's just an issue on my side of if that hasn't been done yet).

One last reminder that some of the components that you're working on might already have pre-made skeletons, so take a look and see if they already exist! That way you won't have to make them yourself! 😄

features/discover/d2l-discover-rules.js
BhamelinD2L: I think one is a better choice as well as they are quite large and it's unlikely to have too many in a standard workflow

remytruong: I've adjusted - realized that the load flow would not allow a situation like above to occur - the checkbox waits for data to be loaded, so it would know the number of rules (or if there are none) by the time the need to render the cards are needed.

features/discover/d2l-discover-rule-card.js
remytruong: On a cursory glance, it doesn't look like the context menu has a skellie, I have elected to hide it!

features/discover/d2l-discover-rule-card.js
remytruong: I wasn't too sure how to skeletonize the match-info section; the load itself for that component requires there to be items existing in the first place, so I elected to skeletonize the whole block. The question in my mind was should I assume that there would be only one user or more in the match-info? It felt jarring when I mocked it to see one user, and some text further away from it, and then see three more boxes load in.

remytruong: In addition, I wanted to wait for the user-profile images to be loaded, but I would have had to add an event to listen for the profile image loading in, which I think would require me to touch the users repo anyway, so I found it to be even more redundant to skeleton out the individual components in match-info.

features/discover/d2l-discover-rule-card.js
remytruong: Additionally, I synced the card title and the match-info section to both finish loading when match-info was complete. That required me to put in an event handler to deskeletize the title when the match-info component loaded all of it's data. I am not sure if that is too smart - I was wrangling with the idea but the title loads so much faster than the match-info component that it looked awful to see.

features/discover/d2l-discover-rule-card-match-info.js
```diff
@@ -68,11 +69,12 @@ class RuleCardMatchInfo extends MatchCountMixin(LocalizeDynamicMixin(RtlMixin(Li
 		super();
 		this._userList = [];
 		this._matchCount = null;
+		this.skeleton = true;
 	}
 
 	render() {
 		return html`
-			<div class="d2l-rule-card-match">
+			<div class="d2l-rule-card-match d2l-skeletize">
```
ChrisLabattD2L, line 77: Instead of combining everything into one loading skeleton, try having separate skeletons for each profile & the count. I've put an example below (though obviously the skeletons aren't properly lined up, which they would need to be).

image

features/discover/d2l-discover-rule-card-match-info.js
```diff
@@ -104,6 +106,12 @@ class RuleCardMatchInfo extends MatchCountMixin(LocalizeDynamicMixin(RtlMixin(Li
 	async _updateMatchInfo() {
 		this._matchCount = this.rule.properties.matchCount;
 		this._userList =  this.rule.properties.userList;
+		this.skeleton = false;
+		const event = new CustomEvent('d2l-rule-card-match-loaded', {
+			bubbles: true,
+			composed: true
+		});
+		this.dispatchEvent(event);
```
ChrisLabattD2L, line 114: These can be combined like, [here's an example](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/features/discover/d2l-discover-rule-picker.js#L379).

features/discover/d2l-discover-rules.js
```diff
@@ -59,6 +59,17 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
...
 				max-width: 25rem;
 			}
+			.d2l-discover-rule-card-skeleton {
+				display: block;
+				border: 1px solid var(--d2l-color-mica);
+				margin-bottom: 0.4rem;
+				max-width: 25rem;
+				min-height: 6rem;
+				background-color:var(--d2l-color-sylvite);
+			}
```
ChrisLabattD2L, line 69: If this class is no longer being used, then it can be removed.

features/discover/d2l-discover-rules.js
```diff
@@ -82,18 +93,17 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
...
 				<div class="d2l-body-small d2l-enrollment-rules">
-					<h5 class="d2l-body-small"><strong>${this.localize('text-rules')}</strong></h5>
-					<p>${this.localize('text-rules-description')}</p>
+					<h5 class="d2l-body-small d2l-skeletize"><strong>${this.localize('text-rules')}</strong></h5>
+					<p class="d2l-skeletize">${this.localize('text-rules-description')}</p>
 					<!-- rules cards -->
-					${this.readOnly ? this._renderReadOnlyRules() : this._renderInteractibleRules()}
+					${ this.readOnly ? this._renderReadOnlyRules() : this._renderInteractibleRules()}
 				</div>
-				` : null}
```
ChrisLabattD2L, line 96: You probably shouldn't remove the conditional render here since the `text-rules` and `text-rules-description` shouldn't be rendered until the user has actually created an enrollment rule. The content that's highlighted below should only appear when a rule exists.

image

ChrisLabattD2L (changes requested):
Some good improvements! But there still a few things that will likely have to be reworked slightly.

features/discover/d2l-discover-rules.js
remytruong: Aaaah right, the conditional render included the above text...

features/discover/d2l-discover-rule-card-match-info.js
remytruong: The only thing I have trouble with is that the match-info component doesn't necessarily know how many profiles there are - the rule-card only manages the title being loaded and the dropdown options menu!

See my comment here: 
https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/347#discussion_r754729115

features/discover/d2l-discover-rules.js
```diff
@@ -68,6 +67,9 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
 				min-height: 6rem;
 				background-color:var(--d2l-color-sylvite);
 			}
+			d2l-labs-checkbox-drawer {
+				max-width: 60%;
```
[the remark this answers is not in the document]
remytruong: Aaaah, I'll set a max-width for the text I want to maintain to be small then!

ChrisLabattD2L (changes requested):
Looks good for the most part, just 2 recommended changes 😄 

I'll come back to this once the changes in the other two repos are finished before this gets approved.

features/discover/d2l-discover-rule-card.js
```diff
@@ -89,14 +89,14 @@ class RuleCard extends DiscoverLocalizationMixin(SkeletonMixin(RtlMixin(LitEleme
 						<d2l-discover-rule-card-match-info @d2l-rule-card-match-loaded="${this._onRuleCardMatchLoaded}" token="${this.token}" .rule="${this.rule}"></d2l-discover-rule-card-match-info>
 					</div>
 				</div>
-				${this.skeleton ? null : html`<d2l-dropdown-context-menu text="options" slot="actions">
+				<d2l-dropdown-context-menu text="options" slot="actions">
```
ChrisLabattD2L, line 92: The user shouldn't be able to access the menu until the content is loading in, so you may still need this since the dropdown button still appears when it's loading in.

remytruong: It just breaks a safari test if I do add this... 
:'( so sad

ChrisLabattD2L: In that case, we'll probably have to find another way to hide it temporarily, maybe adding `class="d2l-skeletize"` to `d2l-dropdown-context-menu` is a potential compromise. I may not look the best, but at least the user can't change anything until it's been loaded in

remytruong: I tried that, but it fails to re-render D:

remytruong: I'll give this one another go - I wanted to avoid touching the d2l-dropdown-context-menu repo though!

BhamelinD2L: What safari test breaks? You might just need to await things to load further in the tests before doing the testing, since it has a newly added delay before the menu exists.

features/discover/d2l-discover-rule-card-match-info.js
BhamelinD2L: I don't mind it all being combined because you're right, there's no way to know what this section will actually look like until we've got at least the profile count. 

The profile images are overlapping each other anyways, so being a list of separate squares isn't representative. Estimating a profile count will cause a jump forward for match count if it's inaccurate. We don't have this issues if we've just got one skeleton, so let's keep it simple.

features/discover/test/d2l-discover-rule-card.test.js
```diff
@@ -36,6 +36,8 @@ describe('d2l-discover-rule-card', () => {
 			el = await fixture(html`
 				<d2l-discover-rule-card .rule="${rule}" token="cake"></d2l-discover-rule-card>
 			`);
+			el.skeleton = false;
+			setTimeout(() => {}, 2000);
```
ChrisLabattD2L, line 40: I probably should've mentioned this earlier, but you should use `await el.updateComplete;` instead of `setTimeout` since you never know how long the timeout will end up needing to be

features/discover/test/d2l-discover-rule-card.test.js
```diff
@@ -53,6 +55,9 @@ describe('d2l-discover-rule-card', () => {
 			el = await fixture(html`
 				<d2l-discover-rule-card .rule="${rule}" token="cake"></d2l-discover-rule-card>
 			`);
+			el.skeleton = false;
+			await el.requestUpdate();
```
ChrisLabattD2L, line 59: I don't think you need `await` here

ChrisLabattD2L (changes requested):
Small improvement for tests that I probably should have mentioned beforehand, sorry about that

features/discover/test/d2l-discover-rule-card.test.js
remytruong: Got it!

ChrisLabattD2L (approved):
The skeletons look good to me now!

I'd maybe recommend @BhamelinD2L take one final look to see if I missed anything here or on https://github.com/BrightspaceUILabs/checkbox-drawer/pull/26

PR #357, opened 2021-11-22
Fix: When reset, conditions now take their original form

ChrisLabattD2L:
Fix for [DE46112](https://rally1.rallydev.com/#/611579333303d/dashboard?detail=%2Fdefect%2F615632406081)

Previously, removing an attribute while editing a pre-existing condition and then clicking on cancel would cause the removed attribute to remain removed if the same rule was opened back up and editing again.

Now, the proper attributes are assigned as it resets the condition to what it was like when it was first being edited.


features/discover/d2l-discover-rule-picker.js
```diff
@@ -265,6 +267,10 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 	}
 
 	reload(newConditions, reloadType) {
+		if (this.ruleIndex !== undefined && reloadType === reloadStatuses.cancel) {
+			this._resetConditionChanges();
```
BhamelinD2L, line 271: This appears to always be running on rule edit, even if done is pressed, causing it to no longer be possible to edit rules.

features/discover/d2l-discover-rule-picker.js
```diff
@@ -671,6 +685,11 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 			skipCondition = true;
 		});
 
+		this._initialConditions = [];
+		this.conditions.forEach((condition) => {
+			this._initialConditions.push(JSON.parse(JSON.stringify(condition)));
+		});
+
```
BhamelinD2L, line 692: The scope of this issue is that if you remove an attribute and press cancel, the attribute remains deleted.
It doesn't affect adding attributes, adding conditions, or removing conditions. These all work correctly in master branch.

Is the scope of duplicating and re-inserting all the conditions actually necessary? How do these other scenarios get handled without this, and can we also apply the logic they use to attribute deletion?

features/discover/d2l-discover-rule-picker.js
ChrisLabattD2L: Good catch, I think it's happening because `_onDoneClick()` wasn't reloading with the `done`, but was still reloading with `cancel` so it was resetting the things to the initial conditions without first submitting the changes. I'll push a fix to this now.

features/discover/d2l-discover-rule-picker.js
ChrisLabattD2L: Unlike the other requested change, this one will likely take longer, so I'll ping if I come up with a simpler or more efficient way of handling this.

ChrisLabattD2L: You also bring up a good point about the fact that adding an attribute and then hitting cancel doesn't affect the conditions the same way that removing one does. I'll start by looking into why that might be.

ChrisLabattD2L: I think I may have figured out the issue, but the change would have to take place in the `multi-select`, I'll make the changes, then link them here.

ChrisLabattD2L: Because of the fix in https://github.com/BrightspaceUILabs/multi-select/pull/123 no changes need to be made in this repo, so I'll close this PR.

ChrisLabattD2L:
Because of the fix in BrightspaceUILabs/multi-select#123 no changes need to be made in this repo, so I'll close this PR.

PR #358, opened 2021-11-22
Fix: Attribute Picker Input Text Reset

ChrisLabattD2L:
Fixes the issue found in [DE46014](https://rally1.rallydev.com/#/611579333303d/dashboard?detail=%2Fdefect%2F614394603929)

Previously if you typed in custom text, then exited the dialog box, then the extra text would continue to appear within the attribute picker input field. Now the text within the attribute picker input field gets removed upon reloading the dialog box.


features/discover/d2l-discover-rule-picker.js
```diff
@@ -268,6 +268,11 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 		this._matchCount = null;
 		this.conditions = [];
 
+		const attributePicker = this.shadowRoot.querySelector('d2l-discover-attribute-picker');
+		if (attributePicker) {
+			attributePicker.emptyInputField();
+		}
+
```
BhamelinD2L, line 275: This is a good place to use `?.` to avoid an if statement.

features/discover/d2l-discover-attribute-picker.js
```diff
@@ -67,6 +67,13 @@ class AttributePicker extends LocalizeDynamicMixin(HypermediaStateMixin(RtlMixin
 		await innerPicker.addAttribute(value);
 	}
 
+	emptyInputField() {
+		const attributePicker = this.shadowRoot.querySelector('d2l-labs-attribute-picker');
+		if (!attributePicker || !attributePicker.shadowRoot.querySelector('input')) return;
+		const inputField = attributePicker.shadowRoot.querySelector('input');
+		inputField.value = '';
+	}
+
```
BhamelinD2L, line 76: `d2l-labs-attribute-picker` has a `clearText()` function in that you can easily simplify this down with, and will keep you from having to dig multiple query selectors deep which should always be avoided if possible. Use `?.` for best results here as well. 

You can't depend on the internals of any other component remaining the same, only the API. Finally, consider naming this clearText as well for naming complexity reduction.

PR #366, opened 2021-11-26
feat: US132870 - Handle Resource Ids

BhamelinD2L:
[US132870](https://rally1.rallydev.com/#/?detail=/userstory/612031414943&fdp=true): [Discover UI]  Do Not Show Resource IDs as User Selectable Values
This PR adjusts conditions to use a name/value object representation in place of a single value representing both.

Tested with live data as well as tests and updated demos.

[This PR requires this branch in the multi-select repo](https://github.com/BrightspaceUILabs/multi-select/pull/118). you can `npm install` this branch into your foundation-component repo, or you can link them both locally with BSI to manually test before merge. Consequently, this PR should have its package version of multi-select updated before merging.


features/discover/mixins/discover-localization-mixin.js
```diff
@@ -11,6 +11,50 @@ export const DiscoverLocalizationMixin = superclass => class extends LocalizeDyn
 		};
 	}
 
+	//Generates a string representation of a given rule to display within cards/rule lists
+	getCardTitle(rule) {
```
BhamelinD2L, line 15: While adjusting the titles to use the .name properties that had been added with this PR, I pulled card title generation into this mixin so that the duplicated code between list and card is removed.

features/discover/test/d2l-discover-rule-picker.test.js
```diff
@@ -399,32 +401,32 @@ describe('d2l-discover-rule-picker', () => {
...
 			el.rules[0]  = {
 				entities: [
-					{ properties: { id:'_fruit', type: 'Fruit', op: '$in' } }
+					{ properties: { id:'_fruit', type: 'Fruit', op: '$in', values: [] } }
 				]
 			};
 			el.rules[1]  = {
 				entities: [
-					{ properties: { id:'_fruit', type: 'Fruit', op: '$in' }, entities: createValueEntities(['apple']) }
+					{ properties: { id:'_fruit', type: 'Fruit', op: '$in', values: [{ name: 'apple', value: 'apple' }] } }
```
ChrisLabattD2L, line 412: I won't comment this for every case, but does the new naming scheme mean that the tests we use should have `name: 'NAME_apple'` instead of just `name: 'apple'`?

ChrisLabattD2L (changes requested):
So I've started going through some of the code, but when I started testing it I was getting the following while using `DiscoverDev`:

image

The cards load in properly though, so it may be an issue with how the attribute picker is being passed the assignable attributes.

I'm gonna continue to look through your code, but thought you may wanna check this out while I do.

features/discover/test/d2l-discover-rule-picker.test.js
BhamelinD2L: > So I've started going through some of the code, but when I started testing it I was getting the following while using `DiscoverDev`:
> 
> image
> 
> The cards load in properly though, so it may be an issue with how the attribute picker is being passed the assignable attributes.
> 
> I'm gonna continue to look through your code, but thought you may wanna check this out while I do.

It did look like this for some period of commits, make sure your instance is fully updated? I'd expect the test to catch that the name on the cards isn't correct. I cleaned out my BSI and re-linked and can't repro this issue.

package.json
```diff
@@ -45,7 +45,7 @@
     "@brightspace-ui-labs/caketray": "^1",
     "@brightspace-ui-labs/checkbox-drawer": "github:BrightspaceUILabs/checkbox-drawer#semver:^1",
     "@brightspace-ui-labs/list-item-accumulator": "^1",
-    "@brightspace-ui-labs/multi-select": "^4.2.4",
+    "@brightspace-ui-labs/multi-select": "git+https://github.com/BrightspaceUILabs/multi-select.git#US132870-named_values",
```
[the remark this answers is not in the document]
BhamelinD2L: Yeah. It'll fail the tests without this, it's noted in the PR request it will be updated.

features/discover/test/test-utils.js
```diff
@@ -6,16 +6,32 @@ export const rels = Object.freeze({
...
+	removed: 'removed',
+	add: 'add',
+	existing: 'existing'
+});
+
+// This can accept either a virgin value or a named value and it will "do the right thing," either way...
+export function createNamedValue(value) {
+	return value.name ? value : {
+		name: `NAME_${value}`,
+		value
```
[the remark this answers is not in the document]
dkong-d2l: If you use the name of a scoped variable as an object name, the value of the variable is used as the value of the object property!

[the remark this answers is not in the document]
BhamelinD2L: We need tests to ensure the name is being used, not the value. Having them be different is useful for that reason.

However, in real data that isn't representing IDs, the name and value match. Making them different doesn't really matter in cases where a difference in name and value isn't what the test data is testing for.

features/discover/test/d2l-discover-rules.test.js
```diff
@@ -340,7 +340,7 @@ describe('d2l-discover-rules', () => {
 
 			const rulePicker = dialog.shadowRoot.querySelector('d2l-discover-rule-picker');
 			const rule = [
-				{ properties: { id:'_orgRole', type: 'Role', values: ['worker'], op: '$in' } }
+				{ properties: { id:'_orgRole', type: 'Role', values: [{ name: 'worker', value: 'worker' }], op: '$in' } }
```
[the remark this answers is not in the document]
BhamelinD2L: It doesn't really matter for the purposes of the test, and this format emulates real data, so I went with the emulation of real data. I don't think it matters in this case, we just need a rule to exist for the test.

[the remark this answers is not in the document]
dkong-d2l: Given that the name of the test is `it(hides add rule when a rule already exists)`, I'm guessing that it does not really matter what the value is called...

features/discover/test/d2l-discover-rule-picker.test.js
ChrisLabattD2L: Deleting and then reinstalling the node modules fixed this issue

features/discover/d2l-discover-rule-picker.js
```diff
@@ -654,8 +642,7 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 				return;
 			}
 
-			const dateVal = (condition.properties.values && condition.properties.values.length > 0) ? condition.properties.values[0] : condition.entities[0].properties.value;
-
+			const dateVal = (condition.properties.values && condition.properties.values[0]) || condition.entities[0].properties.value;
```
dkong-d2l, line 645: This is curious...  Why is `dateVal` sometimes from `condition.properties.values[0]` and sometimes from `condition.entities[0].properties.value`?  Now that we're committing to named values, you'd think it would just be from one source?

BhamelinD2L: I ended up changing this to match the way it behaves from `$in` where it can 
A) take data directly from the siren response as an entity as well as
B) the format passed by the `d2l-discover-rules` component.

There's no reason to handle things via A, as all rules sent to this component are pre formatted by the rule picker to be in line with what is sent out to the BFF on save, so I will remove that entity stuff. Will cause some significant test updates in the picker tests.

features/discover/d2l-discover-rules.js
```diff
@@ -12,7 +12,8 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
 
 const rels = Object.freeze({
 	selfAssignableClass: 'self-assignable',
-	rule: 'https://discovery.brightspace.com/rels/rule'
+	rule: 'https://discovery.brightspace.com/rels/rule',
```
ChrisLabattD2L, line 15: Doesn't have to be done in this PR, but I think it would be worth moving `rels` and other reoccurring constants to a separate file like you did with the [test-utils.js](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/bhamelind2l/us132870-handle-resource-ids/features/discover/test/test-utils.js) file.

ChrisLabattD2L (approved):
Looks good to me, I added a comment regarding moving `rels` and other reused constants to their own file like you had done with the tests, but it doesn't have to be done in this PR. It was more just a clean-up suggestion for the future.

PR #370, opened 2021-12-06
fix: Date Picker value now displays

ChrisLabattD2L:
Context:

- Fixes the defect found in [DE46310](https://rally1.rallydev.com/#/611579333303d/dashboard?detail=%2Fdefect%2F618770060865)
- Because of a change in how values/names are displayed for attributes, the date picker was trying to display an object instead of the actual value.


dkong-d2l:
Is an alternative to this PR to modify the date picker to return just the date string instead of a named value?

ChrisLabattD2L:
> Is an alternative to this PR to modify the date picker to return just the date string instead of a named value?

It could return just the date string, however it chose to keep it in the same format that it would be if it was a pre-existing rule getting loaded in, which would look like this:
```
{
    "name": "2021-11-09",
    "value": "2021-11-09"
}
```

BhamelinD2L (commented):
The code looks clean, but it's introducing another issue where I can't actually open up the calendar when creating a new hire date condition in `npm run start`. There might be an issue with how new dates are being initialized that results it in being considered invalid.

(cant test in discover dev because of the ongoing AWS issues)

ChrisLabattD2L:
> The code looks clean, but it's introducing another issue where I can't actually open up the calendar when creating a new hire date condition in `npm run start`. There might be an issue with how new dates are being initialized that results it in being considered invalid.
> 
> (cant test in discover dev because of the ongoing AWS issues)

I think I know what's causing this, it shouldn't be too difficult to fix. (There's no default value when you switch to the date picker to it causes an error)

I'm also going to take Dave's advice and only pass the value, then convert it into the name & value object in `getSeparatedConditions` to hopefully make it all easier.

dkong-d2l:
The PR seems a little small...  Did you forget to add some changes to your commit?

ChrisLabattD2L:
> The PR seems a little small... Did you forget to add some changes to your commit?

No, I just reverted some of the changes that I had made in the first commit because of your comment about only passing the value instead of both the name and value. As a result, only two things really had to get changed, and things seemed to operate the way they did prior to the defect occurring.

dkong-d2l:
That's funny.  How does this line work?  https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/0b1c2d5fb42fbb0d107a5fa046a3f32c804dae25/features/discover/d2l-discover-rule-picker.js#L594

dkong-d2l (approved):

PR #373, opened 2021-12-08
feat: discover skeletonize attribute picker

tenbradley:
### Context:
[US134410](https://rally1.rallydev.com/#/?detail=/userstory/619025703511&fdp=true)

We're adding skeletons to the rule picker because it's occasionally slow to load and it will improve user experience.

### Quality:
https://user-images.githubusercontent.com/69812595/145272830-a6afabf1-7a47-4c4d-9d91-b83966a20eb3.mp4


features/discover/d2l-discover-rule-picker.js
```diff
@@ -555,16 +561,16 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 				</div>
 				<div class="${classMap(classes)}">
 					${this._onlyOrgRole ? html`
-						<div class="org-role-text ${this._getEmptyConditionClass(condition)} d2l-body-compact">
+						<div class="org-role-text ${this._getEmptyConditionClass(condition)} d2l-body-compact d2l-skeletize">
 							${this.localize('text-role-in-organization')} ${this.localize('text-condition-is')}
 						</div>
 					` : html`
-          				<select class="d2l-input-select d2l-picker-rule-select"
+						<select class="d2l-input-select d2l-picker-rule-select d2l-skeletize"
```
BhamelinD2L, line 568: In your video, the select itself doesn't seem to be skeletized - This might not be working correctly because it is a basic formatted select component, not a d2l component. You might need to do additional work to get this working. Take a look at the docs for this here: https://github.com/BrightspaceUI/core/tree/main/components/skeleton#native-form-elements

features/discover/d2l-discover-rule-picker.js
```diff
@@ -158,22 +159,24 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
...
 
-        		${this._canAddRules() ? html`
+				${this._canAddRules() ? html`
 					<d2l-alert type="default" ?hidden="${!this._isConditionLimitReached()}">${this.localize('alert-condition-limit', 'conditionLimit', this.conditionLimit)}</d2l-alert>
 					<d2l-button-subtle id="add-another-condition-button"
 						text="${this.localize('text-add-another-condition')}"
 						icon="tier1:plus-default"
 						@click="${this._addDefaultCondition}"
-            			?disabled=${this._isConditionLimitReached()}>
+						class="d2l-skeletize"
```
BhamelinD2L, line 176: Could it maybe look better if we added a second css class that shifts the button over to match the margins while being represented as a skeleton? Since the button is subtle and has no background, the fact that it's left shifted to line up the + with the margin makes sense but in the context of loading, I think its positioning looks a little strange.

I'd be interested in what you or others thought on this one.

features/discover/d2l-discover-rule-picker.js
```diff
@@ -584,7 +590,7 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 							href="${this._getConditionTypeHref(condition)}"
 							.token="${this.token}"
 							.attributeList="${condition.properties.values}"
-							class="d2l-picker-rule-attribute-picker"
+							class="d2l-picker-rule-attribute-picker d2l-skeletize"
```
ChrisLabattD2L, line 593: Just in case the first condition type is based on a date, would it also be worthwhile to add the `d2l-skeletize` class to `d2l-discover-date-picker` as well?

features/discover/d2l-discover-rule-picker.js
ChrisLabattD2L: If we don't want to create a class for before it renders, there's also the possibility of hiding the button until everything has been loaded in as well. I don't know if this would lead to any weird rendering issues though.

features/discover/d2l-discover-rule-picker.js
ChrisLabattD2L: Similar to hiding the button, there's the potential of keeping the new skeleton the way it is but also using the `disabled` attribute to prevent the user from using the skeleton until things have loaded in.

tenbradley: Hmm, I assumed with the dropdown being greyed out a bit, that was it skeletonizing the drop down. I'll look into how to a better way to skeletonize the dropdown.

BhamelinD2L: What's your reasoning for considering disabling it while loading over applying the skeleton? I feel like avoiding that state is the reason we have the skeletons in the first place - With the skeleton It's more explicit to the user what's going on and when it's interactable.

features/discover/d2l-discover-rule-picker.js
tenbradley: I agree it looks a bit off. When hovering over the button, the background is consistent for where it actually is. I think just not displaying the button would look better considering it doesn't look like a button once it's loaded.

features/discover/d2l-discover-rule-picker.js
ChrisLabattD2L: At first glance, I thought the blank dropdown was sufficient and just needed to not be interactable. But I can see what you mean about maintaining the consistency and generally avoiding being in that state in the first place.

features/discover/d2l-discover-rule-picker.js
BhamelinD2L: http://design.d2l/patterns/skeleton-loading/
On review we shouldn't be skeletizing it or the `in` separator either:
```
Only dynamic content should be represented by skeleton loading components – static content such as Headings, Buttons, Table Column Headers and Form Controls should not use a skeleton state unless their content is dynamic and depends on external data that takes a long time to load.
```

No dynamic content is actually loading for that button, so it makes sense.
In that case, we can leave the button alone, or put it in a disable state until the dynamic content has finished loading.
In my own testing, mashing it before the dropdown is populated actually works completely fine, so I think it's up to you.

tenbradley:
shinySkeleton
They're prettier

features/discover/d2l-discover-rule-picker.js
BhamelinD2L: From my own testing, it seems the cards won't load in unless the condition dropdowns are also loaded in, likely due to an overlap in the requests to the BFF. I'd recommend double checking in case my network is being funny, but if this is true, you'll never see a skeleton when editing an existing card.

That aside, it's a pretty painless future-proofing, so I'd say you might as well do it. 

I'm also surprised and impressed that this works for `d2l-discover-attribute-picker.js ` given it doesn't [Implement the skeleton mixin](https://github.com/BrightspaceUI/core/tree/main/components/skeleton#skeletizing-custom-elements-with-skeletonmixin) so I guess that means d2l-skeletize works through any descendants that do. That should mean you won't have to do anything special for the date picker either besides slapping `d2l-skeletize` on it, making this pretty simple.

features/discover/d2l-discover-rule-picker.js
```diff
@@ -555,26 +560,28 @@ class RulePicker extends MatchCountMixin(DiscoverLocalizationMixin(HypermediaSta
 				</div>
 				<div class="${classMap(classes)}">
 					${this._onlyOrgRole ? html`
-						<div class="org-role-text ${this._getEmptyConditionClass(condition)} d2l-body-compact">
+						<div class="org-role-text ${this._getEmptyConditionClass(condition)} d2l-body-compact d2l-skeletize">
```
ChrisLabattD2L, line 563: Since the skeleton should only be for dynamic content that loads in, this skeleton could probably also be removed. (Or at least keep it consistent with the `is` text when there's more than one condition type)

tenbradley: ...that snuck in from me trying different things....consider it gone.

ChrisLabattD2L (approved):
Them Skellies are both spooky and scary, and send shivers down your spine

PR #375, opened 2021-12-09
fix: Clabatt/de46350 mobile hired date

ChrisLabattD2L:
Context for [DE46350](https://rally1.rallydev.com/#/611579333303d/dashboard?detail=%2Fdefect%2F619120826783)

For smaller screen sizes, using `in range` for the date picker would cause some overlap with the next condition, so extra padding was added. The padding is only added when the screen's width is small enough to cause the `in range` option of the date picker to appear on multiple lines.

image


BhamelinD2L (commented):
This might be more complicated than necessary. Take a look at discover-fra, which handles this using only css: https://github.com/Brightspace/discovery-fra/blob/8ef09f48514a06b682dd27cecc31c4789c0a2425/src/discovery-settings.js#L161

Have you talked to the designers about this? I don't think we've mentioned anything about mobile development expectations but it would be good to get an approval or some feedback on this change.

ChrisLabattD2L:
> This might be more complicated than necessary. Take a look at discover-fra, which handles this using only css: https://github.com/Brightspace/discovery-fra/blob/8ef09f48514a06b682dd27cecc31c4789c0a2425/src/discovery-settings.js#L161
> 
> Have you talked to the designers about this? I don't think we've mentioned anything about mobile development expectations but it would be good to get an approval or some feedback on this change.

I haven't talked with design about this yet, I was mostly just trying to fix the graphical defect of the overlap. I'll reach out to Jeff about this to see what he has to say about this 👍

ChrisLabattD2L:
@BhamelinD2L we got the go ahead from Jeff 
image

BhamelinD2L:
> @BhamelinD2L we got the go ahead from Jeff
Awesome! Have you looked into the other part of my comment? The javascript might not be necessary if you use `@media only screen and (max-width: ??px)` css. Let me know if that can simplify things.

ChrisLabattD2L:
> > @BhamelinD2L we got the go ahead from Jeff
> 
> Awesome! Have you looked into the other part of my comment? The javascript might not be necessary if you use `@media only screen and (max-width: ??px)` css. Let me know if that can simplify things.

Yeah, so I've tried it, and it does work! However, I would still need to keep a check for if the condition op is `range` since the extra space shouldn't be there for `before` or `after` since those don't overlap. Does that make sense?

BhamelinD2L (approved):
Looks great now. Thank you!

PR #379, opened 2021-12-20
Clabatt/de46357 improve cards

ChrisLabattD2L:
Context for [DE46357](https://rally1.rallydev.com/#/611579333303d/custom/367300408400?detail=%2Fdefect%2F619126301873):

After talking with design, it was determined that we needed to improve what the Rule Cards looked like as they got cluttered and hard to read as more attribute values and conditions are added. So, based on designs found [here](https://miro.com/app/board/o9J_ldQOlT8=/?moveToWidget=3458764515414744092&cot=14), the cards are easier to read.

~~Another addition, though not entirely in the scope of this improvement, adds a subtle button to the cards so that only certain conditions are revealed initially, and by pressing the button the rest of the conditions are revealed.~~

After talking more with Jeff, we've decided not to add the 'load more conditions' button unless clients ask for it, as adding it would create a larger hassle  than it is worth.

Old Design:
image

New Design:
image


features/discover/mixins/discover-localization-mixin.js
```diff
@@ -1,57 +1,73 @@
...
 		return {
 			importFunc: async lang => (await import(`../lang/${lang}.js`)).default
 		};
 	}
 
-	//Generates a string representation of a given rule to display within cards/rule lists
-	getCardTitle(rule) {
-		const conditions = rule.entities;
-		let title = '';
+	getFormattedCardInfo(conditions) {
```
tenbradley, line 24: I'm not a fan of this code being in the localization mixin as it has nothing to do with localization. Ideally we'd have a separate mixin that handles all the Rule data and rendering that both the RuleList and RuleCard extend. However, this is just me caring about single responsibility principle, and this isn't important to change if it would be too much work/doesn't seem like it would be useful to worry about.

This largely occurred to me because I got confused about where this function was coming from and I didn't expect it to be in the localization mixin.

tenbradley (commented):
Aside from my one concern about location of a piece of code, this seems good to me. If we decide not to change those classes, I will approve this.

tenbradley (approved):

PR #385, opened 2022-01-12
US135015 display BONUS when question is bonus

mdulat:
https://rally1.rallydev.com/#/?detail=/userstory/621851497947&fdp=true


emsandrews (approved):
looks good to me!

PR #388, opened 2022-01-25
Non-breaking changes for lit 2

annabelsegalld2l:
This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
- [x] audit `ifDefined` usages for null parameters

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.


ChrisLabattD2L (commented):
So I found a few lit-html imports that were missing `.js`, they all have the following `import { nothing } from 'lit-html';` so I wonder if this can just replaced with with null instead, and reduce the amount of imports. However, adding `.js` is also a possibility.

Aside from that, all of the Discover components work as expect!

annabelsegalld2l:
> So I found a few lit-html imports that were missing `.js`, they all have the following `import { nothing } from 'lit-html';` so I wonder if this can just replaced with with null instead, and reduce the amount of imports. However, adding `.js` is also a possibility.
> 
> Aside from that, all of the Discover components work as expect!

Hey Chris, we don't need to fix the plain `"lit-html"` imports. It would just be the imports for the specific files (ie. `'lit-html/directives/repeat'` becomes `'lit-html/directives/repeat.js'`. The package imports will work correctly. Sorry I didn't specify that in the checklist!

ChrisLabattD2L:
> > So I found a few lit-html imports that were missing `.js`, they all have the following `import { nothing } from 'lit-html';` so I wonder if this can just replaced with with null instead, and reduce the amount of imports. However, adding `.js` is also a possibility.
> > Aside from that, all of the Discover components work as expect!
> 
> Hey Chris, we don't need to fix the plain `"lit-html"` imports. It would just be the imports for the specific files (ie. `'lit-html/directives/repeat'` becomes `'lit-html/directives/repeat.js'`. The package imports will work correctly. Sorry I didn't specify that in the checklist!

Thanks for clearing that up! I'll approve the changes!

ChrisLabattD2L (approved):

PR #394, opened 2022-02-01
Retain checkbox state on rule edit

BhamelinD2L:
[DE47014](https://rally1.rallydev.com/#/?detail=/defect/623596877297&fdp=true): The Discover checkbox is cleared after deleting all Enrollment Rules

This PR resolves the issue where deleting every rule results in the discover checkbox being automatically unchecked from a user perspective.

The cause of this defect was that the check `?checked="${this.isSelfEnrollable || (this._rules && this._rules.length)}"` Would flip as soon as the last rule is deleted, causing a re-render from true->false as an empty checkbox.

There was a property, `_boxUnchecked`, that tracked the checkbox state once the user manipulated it. This variable has been changed to its positive form, `_discoverBoxChecked`, to match the checkbox state. It will now initialize correctly when the rule data loads in addition to being adjusted on user click, so we can use property this to assign checkbox state directly.

I've improved a test to confirm the checkbox state in this scenario.
Manual testing done in Discoverdev to make sure the checkbox saves and loads as expected with different test cases.


tenbradley (approved):
Looks good.

PR #396, opened 2022-02-07
[fix] disable Delete button when question item delete action is not present

stevenao:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fdefect%2F622460647653


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -181,6 +185,18 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 		this.skeleton = !loaded;
 	}
 
+	_checkCanDeleteItems() {
+		const list = this.shadowRoot.querySelector('#quiz-object-list');
+		if (list) {
+			setTimeout(() => {
+				const items = list.getItems();
+				this._canDeleteItems = items.length > 0 && items.every(el => el.deleteAction && el.deleteAction.has);
```
mdulat, line 193: This should probably check for `any` instead of `every` item to have the action. There may be some types in the list that can't be deleted and others that can (like sections, question pools), and I believe we expect that if we end up selecting something that can't be deleted then nothing actually happens. So only if nothing in the list can be deleted should we disable the button.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -181,6 +185,18 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 		this.skeleton = !loaded;
 	}
 
+	_checkCanDeleteItems() {
+		const list = this.shadowRoot.querySelector('#quiz-object-list');
+		if (list) {
+			setTimeout(() => {
```
alexli01, line 191: just curious, why the setTimeout here?

stevenao: It is ugly. It is to run the check after all the question items hypermedia stuff are loaded, otherwise we can't tell if the item has the delete action or not.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
stevenao: It is not simple to disable the button instead of hidding. I spoke to Joseph. We will do hidding for now. As far as the case that some items are having a different delete action state than others, it is probably a non-issue. But as for safety, you point is valid.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -217,6 +221,18 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
...
+				const items = list.getItems();
+				this._canDeleteItems = items.length > 0 && items.any(el => el.deleteAction && el.deleteAction.has);
```
mdulat, line 229: Can we just access `this.items` in the render function to determine if any of them have the action, instead of having this function called from updated?

stevenao: `this.items` - that gives me an idea. I will put it in the `updated` function. Now, it just called once. Putting it in `render()` will result in being check repeatedly.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -180,7 +183,7 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 						<d2l-list id="quiz-object-list" separators="none" @d2l-list-items-move="${this._handleMoveItems}">
 							<d2l-list-header slot="header">
 								${_moreActionsButton}
-								<d2l-selection-action icon="tier1:delete" text="${this.localize('button-quizEditorDelete')}" requires-selection @d2l-selection-action-click="${this._handleDeleteDialogOpen}"></d2l-selection-action>
```
mdulat, line 183: don't we still want the `requires-selection`?

mdulat (approved):

PR #401, opened 2022-02-15
fix: DE46934 - Updated W2D activities to filter on activecoursesonly

PaulKrznaric:
Updated Work To Do to use an alternative route that would cut down on load time experienced by users with elevated permissions to see courses after they are no longer active. 
There should be no visible change to end users but our servers will do less work as a result

Related back end PRs:
Infrastructure Work: https://github.com/Brightspace/lms/pull/17580
Alternate Route Update Work: https://github.com/Brightspace/lms/pull/18837


klhuff (approved):

features/workToDo/d2l-w2d-work-to-do.js
```diff
@@ -21,6 +22,13 @@ const rel = Object.freeze({
 
 class w2dWorkToDo extends LocalizeDynamicMixin(HypermediaStateMixin(LitElement)) {
 	static get properties() {
+		let activitiesToUse;
+		if (D2L.LP && D2L.LP.Web && D2L.LP.Web.UI && D2L.LP.Web.UI.Flags && D2L.LP.Web.UI.Flags.Flag('DE46934-hypermedia-filter-inactive-course-activities', true)) {
```
bearfriend, line 26: Can we use optional chaining here?

lloydwaddell (commented):
LGTM

PaulKrznaric:
@AKobets So I tried to add tests in there but the issue is D2L.LP is invoked somewhere else and is overwritten so I'm unable to mock the Flag within the test. 
I'm also unsure what a test would look like that wasn't somewhat of a test for the sake of a test (i.e. make the method call, confirm you got the correct URL back based on the string)

features/workToDo/d2l-w2d-work-to-do.js
PaulKrznaric: Wasn't aware of that, I'll updat eit now

klhuff (approved):

jordankay13 (approved):

PR #405, opened 2022-02-16
DE47173 - Add deleted labels to unnamed roles and users

BhamelinD2L:
[DE47173](https://rally1.rallydev.com/#/?detail=/defect/626559780179&fdp=true): Discover: Enrollment Rules: When enrollment rule has 'Manager/Org Role' condition set with the user/role which is already deleted, 'undefined' string is displaying as attribute
This PR adds and applies lang terms for deleted users and deleted roles. 
The two new lang terms have been cleared by Jeff. Sample rules in the below screenshots can be checked against in Discoverdev with course id `121223`.

I've included them as a second rule in the main rule test entity, so additional tests were adjusted slightly to match the new rule structure. We didn't have a multi-rule example in the existing tests, so I chose to do it like this instead of making a whole new entity structure for simplicity, and also to make the tests slightly more robust.

image

image

image


features/discover/d2l-discover-rules.js
```diff
@@ -155,6 +155,20 @@ class EntitlementRules extends LocalizeDynamicMixin(SkeletonMixin(HypermediaStat
...
 
+	_addLocalizationToDeletedItems(item) {
+		item.entities.forEach(entity => {
+			entity?.properties?.values.forEach(valueObj => {
+				if (!valueObj.name) {
+					if (entity.properties.id === '_orgRole') {
+						valueObj.name = this.localize('text-deleted-role');
+					} else {
+						valueObj.name = this.localize('text-deleted-user');
+					}
```
BhamelinD2L, line 166: Simple breakdown of the logic being used here:

- Only User and Role types have IDs to pull LMS data from.
- Custom attributes can't have missing names from deletion because they are saved as the name (validated by Nidhi)
- `_orgRole` is the only type of role attribute possible.
Therefore, if the id is not `_orgRole` it must be of type user.

If this statement stops being true at some point we'll need to have the BFF send the attribute type with the rules.

features/discover/test/d2l-discover-rules.test.js
```diff
@@ -272,6 +304,15 @@ describe('d2l-discover-rules', () => {
 			expect(commitSpy.calledWith(expectedCommit), `commit was not called with: ${JSON.stringify(expectedCommit, 2, null)}`).to.be.true;
 		});
 
+		it('creates deleted names for deleted user/role conditions', async () => {
+			const missingNameRule = el._rules[1];
+			const missingUsernameCondition = missingNameRule.entities[0];
+			const missingRolenameCondition = missingNameRule.entities[1];
+
+			expect(missingUsernameCondition.properties.values[0]).to.deep.equal( { value:'11602', name: '-- Deleted User --' } );
+			expect(missingRolenameCondition.properties.values[0]).to.deep.equal( { value:'12345', name: '-- Deleted Role --' } );
```
BhamelinD2L, line 313: The card tests themselves confirm they will render the names and data of cards correctly, so we just need tests to validate the data is here.

tenbradley:
Tested in discover dev and deleted users and roles are being replaced as they should be.

Also noticed that in the roles category, it seems deleted roles might still be present in the list. (Attached video.)

If we want to fix this as a separate defect, I'll approve this.


https://user-images.githubusercontent.com/69812595/154335265-91c1332e-9e40-468e-ae4a-829c8e3ffc49.mp4

BhamelinD2L:
> If we want to fix this as a separate defect, I'll approve this.

Good find - It appears to happen because RoleId 69 for D2LSupport is being returned, which is nameless. I will create a story to block that RoleId from being returned by the BFF, and that should resolve this. Deleted roles won't be returned by the BFF as-is, so that should be sufficient to fix this.

features/discover/test/test-utils.js
```diff
@@ -17,24 +17,24 @@ export const animateOptionsEnum = Object.freeze({
 });
 
 // This can accept either a virgin value or a named value and it will "do the right thing," either way...
-export function createNamedValue(value) {
+export function createNamedValue(value, options) {
 	return value.name ? value : {
-		name: `NAME_${value}`,
+		...(!(options && options.missingNames) && { name: `NAME_${value}` }),
```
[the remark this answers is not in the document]
ChrisLabattD2L: Okay, that makes sense, thanks for clearing it up! Since the defect Ten pointed out is going to be it's own story, I'll approve this to move it along.

ChrisLabattD2L (approved):
Tested it on DiscoverDev for [Ordinary Diff Equations 2](https://discoverdev.devlms.brightspace.com/d2l/home/121223) and it worked as intended. Also going to link [DE47463](https://rally1.rallydev.com/#/357252275780d/defects?detail=%2Fdefect%2F627886477633&fdp=true&fdp=true), so that others that might see that defect will understand that it's in our backlog.

PR #422, opened 2022-02-25
Add oslo to discover localization mixin

tenbradley:
[DE47341](https://rally1.rallydev.com/#/?detail=/defect/627269538065&fdp=true)


features/discover/mixins/discover-localization-mixin.js
```diff
@@ -8,7 +8,8 @@ export const DiscoverLocalizationMixin = superclass => class extends LocalizeDyn
 
 	static get localizeConfig() {
 		return {
-			importFunc: async lang => (await import(`../lang/${lang}.js`)).default
+			importFunc: async lang => (await import(`../lang/${lang}.js`)).default,
+			osloCollection: '@brightspace-hmc\\foundation-components\\DiscoverRules'
```
tenbradley, line 12: This is the only real change in this pr.
Branched from the wrong branch, so the other updates are actually in master already.

BhamelinD2L (approved):

PR #426, opened 2022-03-01
Intro PR for automatic grading tag component 

mattkraljevic:
Created lit element component
image
changes will be made after hypermedia work is done


features/autoZero/lang/en.js
```diff
@@ -0,0 +1,4 @@
+/* eslint quotes: 0 */
+export default {
+	"auto-zero-message": "Automatic zero. Missing submission on the due date."
```
DiljotSG, line 3: We might want to verify this terminology with @nicolebezaire first

features/autoZero/d2l-auto-zero-grade.js
```diff
@@ -0,0 +1,50 @@
+import '@brightspace-ui/core/components/tooltip/tooltip.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { HypermediaStateMixin, observableTypes } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
+
+const rels = {
+	desiredLinkRel: 'self'
```
apalaniuk, line 7: nit: I'd just inline this, since `self` provides the relevant semantic context already.

features/autoZero/d2l-auto-zero-grade.js
```diff
@@ -0,0 +1,50 @@
...
+		return css`
+			.auto-zero-tooltip {
+				text-align: center
+			}
+			.d2l-auto-zero-active {
+				display: inline-block;
+				text-decoration: underline dotted;
+			}
+			.d2l-auto-zero-inactive {
+				display: inline-block;
```
apalaniuk, line 27: You could put this under a `:host` selector, then remove this case, and the `display` on the active case

features/autoZero/d2l-auto-zero-grade.js
```diff
@@ -0,0 +1,50 @@
...
+	static get properties() {
+		return {
+			isAutoZeroed: { type: Boolean, observable: observableTypes.property, name: 'isAutoZeroed', route: [{observable: observableTypes.property, rel: rels.desiredLinkRel}] },
+		};
+	}
+
+	static get styles() {
+		return css`
+			.auto-zero-tooltip {
+				text-align: center
```
apalaniuk, line 20: nit: 
```suggestion
				text-align: center
```
```suggestion
				text-align: center;
```

(ideally, this would be linted by stylelint)

features/autoZero/d2l-auto-zero-grade.js
```diff
@@ -0,0 +1,50 @@
...
+
+	static get localizeConfig() {
+		return {
+			importFunc: async lang => (await import(`./lang/${lang}.js`)).default
+		};
+	}
+
+	constructor() {
+		super();
+		this.test = true;
```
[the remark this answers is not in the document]
mattkraljevic: image
I'm getting this error after I remove the line, I'm just going to remove the entire constructor for now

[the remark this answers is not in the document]
apalaniuk: You can omit the constructor if it would otherwise be empty, as default constructors of derived classes will just call the parent class constructor (`super()`).

features/autoZero/lang/en.js
mattkraljevic: I'll touch base

features/autoZero/d2l-auto-zero-grade.js
mattkraljevic: I believe I did the :host syntax right - let me know if there's a mistake in it.

features/autoZero/lang/en.js
mattkraljevic: just got word this is the correct terminology

features/autoZero/demo/data/auto-zero-false.json
```diff
@@ -0,0 +1,9 @@
+{
+  "class": ["autoZeroResult"],
+  "properties": {
+    "isAutoZeroed": false
+  },
+  "links": [
+    { "rel": ["self"], "href": "auto-zero.json"}
```
apalaniuk, line 7: (not a huge deal because this won't _do_ anything functional in the demo, but this file doesn't exist, which might break the demo later. Could just reference its own file in both cases (true/false).)

mattkraljevic: ah whoops. This was the old name of the file, I'll update it to the current name.

imcdonald-d2l:
Moving work to my own branch while I look into integrating this into the grades page: https://github.com/BrightspaceHypermediaComponents/foundation-components/tree/imcdonald/US138981-grade-annotate

PR #427, opened 2022-03-01
us136458-support-hindi

BhamelinD2L:
Hindi was previously added with another localization PR. This PR adds the hi.js to the static localization file list for each of the different components.

Remaining localization mixins such as the ones for the Discover feature are dynamic and don't require updating.

`fr-on.js` has also been added to the quit component, as the import was missing. The file is identical to the `fr` version, so this should be safe.


ChrisLabattD2L (approved):
Good catch adding in `fr-on` when it was missing, the rest looks good too!

PR #438, opened 2022-03-14
US136651: create hypermedia filter component

tenbradley:
### Context
[US136651](https://rally1.rallydev.com/#/?detail=/userstory/627974778725&fdp=true)

After chatting with Brad, realized that this should not be a hypermedia component as we don't have an route that we'll retrieve data from directly for a filter. The filter controller will retrieve the data and send it to each filter. Thus, I have made this component not be a hypermedia component.

We still want this non-core filter as we've added the ability to update filter values while maintaining whether values have been selected. This will be needed when we get to dynamic filter values.


components/common/d2l-hc-filter.js
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	_onFilterChange(e) {
+		let updated = false;
+		if (e.detail.dimensions.cleared) {
+			this.selected.clear();
+			updated = true;
+		} else {
+			e.detail.dimensions[0].changes.forEach(element => {
+				if (this.selected.get(element.valueKey) !== element.selected) updated = true;
```
ChrisLabattD2L, line 49: This could easily just be `updated = (this.selected.get(element.valueKey) !== element.selected)`, you don't really need the if

components/common/d2l-hc-filter.js
```diff
@@ -0,0 +1,64 @@
...
+			href: { type: String, reflect: true },
+			title: { type: String, observable: observableTypes.property },
+			values: { type: Object, observable: observableTypes.property }
+		};
+	}
+
+	constructor() {
+		super();
+		this.title = undefined;
+		this.values = undefined;
```
ChrisLabattD2L, line 21: You don't actually have to initialize title and values to be `undefined` since they will already be `undefined` at this point.

components/common/d2l-hc-filter.js
```diff
@@ -0,0 +1,64 @@
...
+		this.selected = new Map();
+	}
+
+	render() {
+		if (!this.title || !this.values) return html``;
```
ChrisLabattD2L, line 26: You don't actually have to `return html'';` here, you can just do `return;`

components/common/test/d2l-hc-filter.test.js
```diff
@@ -0,0 +1,167 @@
...
+};
+
+const threeCourses = [course1, course2, course3];
+// const responseThreeItems = {
+// 	'class': ['filter-type'],
+// 	'properties': {
+// 		'title': dimension1,
+// 		'values': threeCourses
+// 	}
+// };
```
ChrisLabattD2L, line 41: If we don't end up needing this, you might as well remove it, it isn't hard to readd it later if it does end up being needed.

ChrisLabattD2L (changes requested):
Pretty much just some nitpicks from my end.

components/common/d2l-hc-filter.js
tenbradley: Changed to `updated = updated || (this.selected.get(element.valueKey) !== element.selected)`

components/common/d2l-hc-filter.js
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	_onFilterChange(e) {
+		let updated = false;
+		if (e.detail.dimensions.cleared) {
+			this.selected.clear();
+			updated = true;
+		} else {
+			e.detail.dimensions[0].changes.forEach(element => {
+				updated = updated || this.selected.get(element.valueKey) !== element.selected;
```
ChrisLabattD2L, line 46: Since you set updated to false just before this, I don't think you'll actually need the `updated ||` part

components/common/test/d2l-hc-filter.test.js
```diff
@@ -0,0 +1,134 @@
+/* eslint-disable no-undef */
+import '../d2l-hc-filter.js';
+import '@brightspace-ui/core/components/filter/filter.js';
+import '@brightspace-ui/core/components/filter/filter-dimension-set.js';
+import '@brightspace-ui/core/components/filter/filter-dimension-set-value.js';
+import { expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { clearStore } from '@brightspace-hmc/foundation-engine/state/HypermediaState.js';
```
ChrisLabattD2L, line 7: Since the hypermedia aspect was removed, I don't know if we still need to use `clearStore`

ChrisLabattD2L (changes requested):
A few little nitpicks, but I did have one more important question which is, what happens if we have more than one value with the same name? Is this a situation that we should be dealing with? Or are we going under the assumption that there'll never be a situation in which this happens?

components/common/d2l-hc-filter.js
```diff
@@ -0,0 +1,61 @@
+import '@brightspace-ui/core/components/filter/filter.js';
+import '@brightspace-ui/core/components/filter/filter-dimension-set.js';
+import '@brightspace-ui/core/components/filter/filter-dimension-set-value.js';
+
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { HypermediaStateMixin } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
```
ChrisLabattD2L, line 6: Also forgot to ask, do we still need this mixin?

components/common/d2l-hc-filter.js
tenbradley: There could be multiple changes. The `updated ||` should account for if there are multiple items in the changes array but some are unchanged.

components/common/d2l-hc-filter.js
tenbradley: I'm thinking about this. I think it should be safe to remove this. I'll try that out.

components/common/d2l-hc-filter.js
ChrisLabattD2L: You're absolutely correct, forgot it was in a loop oops

components/common/test/d2l-hc-filter.test.js
```diff
@@ -0,0 +1,134 @@
...
+			expect(elementSpy.calledOnce, 'hc filter does not emit event').to.be.true;
+
+			expect(filterSpy.called, 'the d2l-filter did not emit a change filter event').to.be.true;
+			expect(element.selected.has(course1.name), 'department is not selected in hm filter').to.be.true;
+		});
+
+		it('should update filter values', async() => {
+			const element = await _createComponent();
+			expect(element.values.length, 'filter component is not initialized with two values').to.equal(2);
+			element.updateValues(threeCourses);
```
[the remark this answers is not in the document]
tenbradley: I potentially could get rid of the updateValues function. Or at the very least change it to a settor. While it's not being used for it currently, I forsee this function being updated to handle dynamic filters. But based on it's use now, I can get rid of it.

ChrisLabattD2L (approved):
Looks good! Thanks for making those changes!

PR #440, opened 2022-03-15
Default `filter-inactive-course-activities` flag to false

bearfriend:
[DS13461](https://rally1.rallydev.com/#/?detail=/defectsuite/629353479767&fdp=true): Work to Do Widget > 20.22.2.35263 > The Work To Do Widget is not displaying upcoming or overdue activities. Work to Do Widget shows All clear at Org Level for Assignment due in 1 week


klhuff:
I don't see any harm in this change, but it is not currently necessary as for a hotfix, we're going to backport the flag itself into 20.22.3.

mbardakovD2L (approved):

PaulKrznaric:
I think @klhuff was thinking we could do both

klhuff:
> I think @klhuff was thinking we could do both

We can, but it's not necessary. Getting just this change into a release branch of this repo is much harder than us just making a change to the LMS.

PR #445, opened 2022-03-16
fix: work to do bullet

tenbradley:
### Context
The bullet is being read as an image by screen readers in the brightspace for parents app. We believe this is due to an svg being used for this bullet.

This pr changes the bullet to the UFT bullet and updates the spacing because this bullet is smaller than the svg was.

### Quality:
Tested in work to do in sitelord. Have not tested with a screen reader.


mbardakovD2L:
I tested this on my local machine with VoiceOver and the bullet is read out literally as "bullet."

mbardakovD2L (approved):

PR #448, opened 2022-03-16
Eandrews/us135066 set points UI

emsandrews:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F621961024647

Implementing UI for set Points


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -376,6 +400,40 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
...
+			//convert to string
+			this.setPointsValue = points.toString();
+			//add to action state
+			this.actionStatePoints = this.actionStatePoints.concat(this.setPointsValue);
+			//encode action state and update
+			const actionStateEncoded = window.btoa(this.actionStatePoints);
+			this.bulkSetPointsAction.commit({actionStates: actionStateEncoded});
+			await this._state.push();
+			this._setPointsDialogOpened = false;
+			await this._refreshState();
```
mdulat, line 424: Do we have to refresh the entire collection here? We're already refreshing the updated items below.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -169,14 +186,20 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
...
+		this._setPointsValue = '';
+		this._actionStatePoints = '';
 
 		this.addEventListener('d2l-question-updated', this._handleQuestionUpdate);
 		this.addEventListener('d2l-collection-item-delete-dialog-confirm', this._handleDeleteDialogConfirm);
 		this.addEventListener('d2l-collection-item-delete-dialog-cancel', this._handleDeleteDialogCancel);
 		this.addEventListener('d2l-activity-move-to-section', this._handleMoveToSection);
+		this.addEventListener('d2l-collection-item-set-points-dialog-confirm', this._handleSetPointsDialogConfirm);
+		this.addEventListener('d2l-collection-item-set-points-dialog-cancel', this._handleSetPointsDialogCancel);
+		this.addEventListener('d2l-collection-item-reset-points-dialog', this._resetPointsDialog);
```
mdulat, line 202: Do we need a separate event `d2l-collection-item-reset-points-dialog`, isn't it handled the same as `d2l-collection-item-set-points-dialog-cancel`?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	_clearInputField() {
+		const input = this.shadowRoot.querySelector('#set-points-input');
+		input.value = null;
+	}
+
+	_handleDialogClose() {
+		//if points are valid, or if cancel is clicked, or if the "X" is clicked, then close the dialog
+		if (this._pointsValid || this._cancelDialog || !(this._cancelDialog && this._confirmDialog)) {
```
mdulat, line 80: can `this._cancelDialog` && `this._confirmDialog` both be true at the same time?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
```diff
@@ -0,0 +1,158 @@
...
+				this.dispatchEvent(new CustomEvent('d2l-collection-item-reset-points-dialog', {bubbles: true, composed: true,
+					detail: {
+						resetDialog: true
+					},
+				}
+				));
+			}
+
+			//close the dialog
+			this.dialogOpened = false;
```
mdulat, line 93: I'm confused as to why we're managing the `dialogOpened` both in this component and the parent component.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
```diff
@@ -0,0 +1,158 @@
...
+
+			//reset all local variables
+			this._pointsValue = null;
+			this._clearInputField();
+			this._confirmDialog = false;
+			this._cancelDialog = false;
+
+		} else {
+			//keep the dialog open
+			this.dialogOpened = true;
```
mdulat, line 103: Is this necessary, wouldn't `this.dialogOpened` still be true at this point?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	_isInputPointsValueValid(points) {
+		//check if points value input is valid
+		return !(points === undefined || points <= (minPoints - 1) || points >= (maxPoints + 1));
```
mdulat, line 109: I think you're assuming the input will be an integer. What if the user inputs 0.5?
Also I find it confusing to negate the whole thing, maybe that's just me. Would it be simpler to just list the conditions that make valid true? I.e. `points !== undefined && points >= minPoints && points <= maxPoints`

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
```diff
@@ -0,0 +1,158 @@
...
+		this.dispatchEvent(new CustomEvent('d2l-collection-item-set-points-dialog-cancel', {bubbles: true, composed: true}));
+
+		//reset points value and input for next open
+		this._pointsValue = null;
+		this._clearInputField();
+	}
+
+	_onSetPointsSelectedContent() {
+		//if the points value is valid, proceed with confirm
+		if (this._pointsValid) {
```
mdulat, line 126: I think you can just check validity here instead of keeping track of it every time the use inputs it. The component does the validity checking itself so you don't have to handle it until the user clicks Save. Then all you have to do is not send the confirm event if it's invalid. When you do send the event, it will be valid so you don't have to pass the `pointsValid` value because the parent can assume that if it gets the event then it is valid.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
emsandrews: good point, I updated it to use the `d2l-dialog-close` event to signal when the dialog is closed and then call the same `d2l-collection-item-set-points-dialog-cancel` event.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -376,6 +400,40 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 		this._refreshNumberingState();
 	}
 
+	_handleSetPointsDialogCancel() {
+		//if cancel is triggered close dialog
+		this._setPointsDialogOpened = false;
+		this._state.reset();
```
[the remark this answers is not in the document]
emsandrews: so I realized I don't need this. This is supposed to refresh the state in the case that the state is pushed, but then the user changes their mind and cancels. In this case, the state doesn't get pushed unless a valid points value is entered.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
emsandrews: you are right, `await this._refreshState();` is not needed here

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
emsandrews: I removed this when I updated how cancelling with the "X" is handled.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
emsandrews: I removed this

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -169,14 +186,19 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 		this.skeleton = true;
 		this.numbering = undefined;
 		this._deleteDialogOpened = false;
+		this._setPointsDialogOpened = false;
 		this._refreshAllCounter = 0;
 		this._highlightCounter = 0;
 		this.selectedItems = [];
+		this._setPointsValue = '';
```
mdulat, line 193: Do you still need this variable?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -376,6 +399,38 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
...
+	}
+
+	async _handleSetPointsDialogConfirm(e) {
+
+		//if points value input is vaid
+		if (e.detail.points) {
+			//get points value
+			const points = e.detail.points;
+			//convert to string
+			this.setPointsValue = points.toString();
```
mdulat, line 414: I think this can be a local variable

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
```diff
@@ -0,0 +1,122 @@
...
+import '@brightspace-ui/core/components/dialog/dialog.js';
+import '@brightspace-ui/core/components/inputs/input-number.js';
+import { css, LitElement } from 'lit-element/lit-element.js';
+import { html } from '@brightspace-hmc/foundation-engine/framework/lit/hypermedia-components.js';
+import { HypermediaStateMixin } from '@brightspace-hmc/foundation-engine/framework/lit/HypermediaStateMixin.js';
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
+
+const maxPoints = 10000;
+const minPoints = 0;
```
mdulat, line 11: I don't think this is right. If you're setting minPoints to 0 exclusively, then 0.5 is a valid input. Similarly for maxPoints.

emsandrews: I believe that 0.5 should be a valid input for points. At least, this is the behaviour in QED.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
emsandrews: nope

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-set-points-quiz.js
mdulat: Ok, that's different from the story description so maybe double-check with Joseph?

mdulat (approved):

PR #450, opened 2022-03-21
fix: dependencies

devpow112:
Causing BSI failures https://github.com/Brightspace/brightspace-integration/pull/6019.


devpow112:
NM related to lit 20 work

PR #453, opened 2022-03-23
feat: filter controller

tenbradley:


components/common/d2l-hc-filter.js
```diff
@@ -21,16 +21,25 @@ class HmFilter extends LitElement {
 	render() {
 		if (!this.title || !this.values) return;
 		return html`
+		<button @click="${this.updateOne}">Update</button>
 			<d2l-filter @d2l-filter-change="${this._onFilterChange}">
 				<d2l-filter-dimension-set key="${this.title}" text=${this.title}>
-					${repeat(this.values, value => html`
+					${repeat(this.values, (value) => value.orgUnitId, value => html`
```
AKobets, line 27: why this component is specific to orgUnitId? How we are going to reuse it for other types of data ?

components/common/d2l-hc-filter.js
```diff
@@ -21,16 +21,25 @@ class HmFilter extends LitElement {
...
 						<d2l-filter-dimension-set-value key="${value.orgUnitId}" text="${value.name}" ?selected="${this.selected.get(value.orgUnitId)}"></d2l-filter-dimension-set-value>
 					`)}
 				</d2l-filter-dimension-set>
 			</d2l-filter>
 		`;
 	}
 
+	updateOne() {
```
AKobets, line 35: should this method have a comment explaining why it has hardcoded data ?

components/common/test/d2l-hc-filter-controller.test.js
```diff
@@ -0,0 +1,264 @@
...
+	describe('construction', () => {
+
+		it('should construct', () => {
+			// use the core runConstructor to ensure tag exists
+			runConstructor('d2l-hc-filter-controller');
+		});
+	});
+
+	describe('accessibility', () => {
+		it('should pass all aXe tests', async() => {
```
AKobets, line 129: empty component states are not good enough for axe tests

components/common/d2l-hc-filter.js
AKobets: would it be better to have value.key or value.data instead of value.orgUnitId ?

components/common/test/d2l-hc-filter-controller.test.js
```diff
@@ -0,0 +1,264 @@
...
+			// open the filter and wait for the first open event to fire
+			const hcfilter = element.shadowRoot.querySelector('d2l-hc-filter');
+			const filter = hcfilter.shadowRoot.querySelector('d2l-filter');
+			filter.opened = true;
+			await oneEvent(filter, 'd2l-filter-dimension-first-open');
+
+			expect(element.values/*.has(course1.orgUnitId)*/, 'selected values intialized before selection').to.be.undefined;
+
+			// check the math option in the filter list and wait for the corresponding change to fire an event
+			const mathItem = filter.shadowRoot.querySelector('d2l-list-item');
```
AKobets, line 193: what is  a trick with name. how do you know it is a mathItem ?

components/common/test/d2l-hc-filter-controller.test.js
```diff
@@ -0,0 +1,264 @@
...
+
+			expect(elementSpy.calledOnce, 'hc filter controller does not emit event').to.be.true;
+			const eventDetails = elementSpy.getCalls(0)[0].args[0].detail;
+			expect(eventDetails, 'filter controller event details does not match selected values').to.deep.equal(element.values);
+
+			mathItem.setSelected(false);
+			await oneEvent(filter, 'd2l-filter-change');
+
+			expect(element.values[0].selected.size, 'first filter should have no selected values').to.equal(0);
+			expect(elementSpy.calledTwice, 'hc filter controller does not emit event').to.be.true;
```
AKobets, line 218: maybe `elementSpy` is better to rename to `dispatchEventSpy`.
Is there way to assert actual arguments for spy call?

components/common/test/d2l-hc-filter-controller.test.js
```diff
@@ -0,0 +1,264 @@
...
+
+			expect(element.filters, 'controller did not initialize filters array').to.be.instanceOf(Array);
+			expect(element.filters.length, 'filters array should have two filters').to.equal(2);
+			expect(element.filters[0].properties.values, 'first filter should have no values').to.deep.equal([]);
+			expect(element.filters[1].properties.values, 'second filter should have no values').to.deep.equal([]);
+
+			element.updateFilterValues([departmentEntity.properties, emptySemester.properties]);
+			await element.updateComplete;
+
+			expect(element.filters[0].properties.values.length, 'first filter should have no values').to.equal(3);
```
AKobets, line 232: incorrect assertion message. should be `first filter should have 3 values`

components/common/d2l-hc-filter-controller.js
```diff
@@ -0,0 +1,80 @@
...
+
+	selectedFiltersUpdated(e) {
+		// if this is the first time we've updated, initialize the values
+		if (!this.values) {
+			this.values = [];
+			this.filters.forEach(dimension => {
+				this.values.push({dimension: dimension.properties.title, selected: new Set()});
+			});
+		}
+		const filterTitle = e.detail.dimensions[0].dimensionKey;
```
AKobets, line 34: is it a bug? - filterTitle is assigned value from the first dimension, however values are pushed for all dimension

components/common/d2l-hc-filter-controller.js
```diff
@@ -0,0 +1,80 @@
...
+			this.dispatchEvent(new CustomEvent('d2l-hc-selected-filters-updated', {
+				bubbles: true,
+				composed: false,
+				detail: this.values
+			}));
+		}
+	}
+
+	updateFilterValues(newValues) {
+		// [{title: name, values: [{name: "name", orgUnitId: "1234"}]}, ...]
```
AKobets, line 68: > can we ensure that the filters are in the same order as we originally got them in?

may be implement filters as map and not an array object ?

components/common/d2l-hc-filter.js
tenbradley: Forgot to remove data from testing....

components/common/d2l-hc-filter-controller.js
tenbradley: I originally had it as a map. I changed it to an array as we need to have some way of knowing which filter gets each set of values upon being updated. I figured since we send data in a certain order, it wouldn't be hard for bff to return it in the same order. I'll think about how to do it with a map.

components/common/d2l-hc-filter.js
tenbradley: The `(value) => value.orgUnitId` was added to fix an issue arising from lit trying to be smart. According to Stacey, lit tries to reuse dom nodes to reduce unnecessary rerendering. When the values were being updated, it was occasionally not rendering them correctly when a value for a filter was being overridden with a new value. My understanding is that this forces it to rerender if the orgUnitId has changed.

components/common/test/d2l-hc-filter-controller.test.js
tenbradley: I know it's math based on the data given to the component. Query selector will return the first instance of a d2l-list-item, and math is the first one. If you prefer, I can change this to be more explicit.

components/common/d2l-hc-filter-controller.js
tenbradley: We are using multiple single dimension filters. I am only using the first dimension here as it's the only one that will ever exist.

components/common/test/d2l-hc-filter-controller.test.js
AKobets: yes, please change it to firstItem

components/common/d2l-hc-filter-controller.js
tenbradley: Comment added

components/common/test/d2l-hc-filter-controller.test.js
tenbradley: Added an aXe test for a controller with two filters.

AKobets (approved):

PR #454, opened 2022-03-23
fix: DE47654 label checkboxes with names instead of ids

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/628618548045&fdp=true


components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -117,6 +117,10 @@ const componentClass = class extends HypermediaStateMixin(ListItemButtonMixin(Li
 			},
 			highlightCounter: {
 				type: Number
+			},
+			label: { type: String, reflect: true },
```
emsandrews, line 121: where is this label property used?

mdulat: On the checkbox of the list item, so screen reader will read the label to indicate what the checkbox is for.

emsandrews (approved):

PR #473, opened 2022-04-01
more robust filter and adding a larger filter controller demo

tenbradley:
### Context:
Updated controller code to reset selected filter values based on the new set of values.
Previously, if a values was selected and the filter was updated to not contain the value, the old selected filter that no longer exist were still selected.
This was fine since we never expect values to be removed from a filter if they are selected in our use case. However, since this is a generic hypermedia component, I decided to add this robustness because who knows how others might want to use this.

Added a filter controller to the demo that uses more filters just to show off that it's possible.


ChrisLabattD2L (commented):
Should the bigger test maybe update the `Month` and `Type` filters when clicking on the `Simulate Update` button in addition to updating `Department` and `Semester`? Just so that the demo is more thorough?

tenbradley:
> Should the bigger test maybe update the `Month` and `Type` filters when clicking on the `Simulate Update` button in addition to updating `Department` and `Semester`? Just so that the demo is more thorough?

It could. I wanted to add more filters mostly to show that more filters can be displayed and all the functionality is really displayed in the second demo. But I'm not opposed to updating them in some way.

tenbradley:
> New demo looks good!
> 
> I'm approving, but there's a test case that I think might be worth addressing at some point. If an option is selected, then gets removed and re-added, should it be selected when it gets added back, or no? I don't know if this has been talked about yet or not.

This is actually the last test case I added to the filter itself. I figured it should not be selected if it is added back, but I could see it being either.

ChrisLabattD2L:
> > New demo looks good!
> > I'm approving, but there's a test case that I think might be worth addressing at some point. If an option is selected, then gets removed and re-added, should it be selected when it gets added back, or no? I don't know if this has been talked about yet or not.
> 
> This is actually the last test case I added to the filter itself. I figured it should not be selected if it is added back, but I could see it being either.

As long as it has been noted, then I'm okay with it not being selected when re-added. Worse case scenario, we add that in later.

PR #478, opened 2022-04-04
fix: US137145 maintain keyboard focus when moving list items into, ou…

mdulat:
…t of, and within sections

https://rally1.rallydev.com/#/?detail=/userstory/629729202029&fdp=true

The problem is that when list items are moved into or out of sections, they are re-constructed. The `list-item-drag-handle` component sets focus on an element that has just been moved. However that element is disconnected and a new one takes its place. Even if I set focus on the new element when it's constructed, it loses focus when the old element gets focus, then when the new element gets focus back it is not in keyboardActive mode (because that was set to false when it lost focus). So I have to activate the keyboard and set focus after first update, then do it again after second update (to account for the times when it looses focus to the old element after I set it the first time).

This seems like a very hacky solution but I've been fighting with this for most of last week. If anyone has a better approach I'd be very open to suggestions!


mdulat:
> Something to note is that the brightspaceUI demo page also loses focus when moving into sections, so trying to fix it in the core repo could be a cleaner solution.

I'm not seeing the demo page lose focus when moving into sections. On which page did you see that? I'm looking at "d2l-list (with drag & drop)", "Move Event" section.

austinkwok1997:
I was looking at the same section but I think I was on an outdated version of brightspace-ui/core. It doesn't lose focus for me anymore.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
```diff
@@ -279,6 +295,10 @@ const componentClass = class extends HypermediaStateMixin(ListItemButtonMixin(Li
 		if (changedProperties.has('_keyboardActive') && changedProperties.has('_loaded')) {
 			this.activateDragHandle();
 		}
+		if (changedProperties.has('_focusingDragHandle') && this._focusMe) {
+			this._focusKeyboardHandle();
```
annabelsegalld2l, line 299: Seems like if _focusMe is set to true, then we have already run _focusKeyboardHandle(). Do we need to run it here again?

annabelsegalld2l (commented):
I can understand the solution here, super job Margaret for figuring this out! I think it would be helpful to have some extra comments for future us - it will also be helpful if this bug ever gets fixed by internal list changes, then we can remove these properties.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js
annabelsegalld2l: Ah, missed your explanation in the summary!

dbatiste:
> I was looking at the same section but I think I was on an outdated version of brightspace-ui/core. It doesn't lose focus for me anymore.

Previously the demo code was using `requestAnimationFrame` to wait for things to be rendered, but it wasn't sufficient.  It was recently updated to use a `setTimeout(, 0)` instead, which might be why you are unable to reproduce what you originally observed in the core demo.  In any case, the item updates (resulting from the move) need to be complete and rendered before trying to activate or focus them.  

I have some idea on how to make this aspect of the API better but I need to wrap up some other things first.

alexli01 (approved):

PR #481, opened 2022-04-08
fix: add null checking to avoid console errors

mdulat:
Sometimes moving items caused console errors `Cannot read properties of null` when `list` or `keyboardText` was null. So added null checks.
Also moved the `list?.getItems()` call out so that it's only made once instead of for every item.

I'm not sure what other effects there are when `list` or `keyboardText` is null, will have to do more testing on quad once this is in.


dustxd (approved):

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -667,20 +667,22 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
...
 
-			if (element.value.key === targetItem.key) {
-				keyboardText.currentPosition = element.position;
+			if (element?.value?.key === targetItem?.key) {
+				keyboardText.currentPosition = element?.position;
 				// keyboardText.count isn't checked as it doesn't seem to be out of sync
 				break;
 			}
 			if (elementIsCollection) {
 				const elementListItem = element.value.shadowRoot?.querySelector('d2l-list-item-generic-layout');
```
dustxd, line 682: Do we need to check if `element.value` exists too?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -667,20 +667,22 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 		const listCollection = [];
 		const keyboardText = {...oldKeyboardText};
 
-		list.getItems().forEach((item, idx) => listCollection.push({value: item, parentSize: list.getItems().length, position: idx + 1}));
+		const listItems = list?.getItems();
+		listItems.forEach((item, idx) => listCollection.push({value: item, parentSize: listItems.length, position: idx + 1}));
```
dustxd, line 671: Just curious based on the PR description - would this make a performance difference? I thought `list.getItems().forEach()` would mean `list.getItems()` gets called once and then iterated on., but from the PR description it sounds like `list.getItems()` would get called n times?

mdulat: You are correct that list.getItems() gets called once and then iterated on, however the extra calls were actually within the foreach function where it was doing `list.getItems().length` for each item.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
mdulat: yeah, wouldn't hurt to add that. I just assumed that since this was within the conditional `elementIsCollection` then `element.value` would have to exist. But that's a bad assumption since in future code may get shuffled around. I'll add it.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
dustxd: Ohhh I see it now. In that case should we move the call to `.length` out as well?

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
dustxd: I see - the same check was done [here](https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/481/files#diff-d8c2c48554b226e13dce6bb64f7fae1385da0cf2c8f2fb41627df050908de769R674). Thanks for adding that anyway!

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
mdulat: hmm, that's ones supposed to be constant time as it's a property on the array, so I think it's less important. Like it's not computed each time you call it.

dustxd: Oooo good to know 👍

dustxd (approved):

PR #486, opened 2022-04-12
fix: DE48056 highlight item after drag & drop via mouse

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/631137836143&fdp=true


mostlyfabulous (approved):
Is the delay of 500ms based on the current responsiveness of the UI?

mdulat:
> Is the delay of 500ms based on the current responsiveness of the UI?

Yes unfortunately it's not ideal. If we leave it at `0` then sometimes it works, and sometimes it's a little too soon.

PR #491, opened 2022-04-12
DE48024 section menu keyboard fix

austinkwok1997:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F630933450469


components/activity/editor/collection/custom/quiz/d2l-activity-collection-section-menu-item.js
```diff
@@ -47,12 +47,22 @@ class ActivityCollectionSectionMenuItem extends SkeletonMixin(HypermediaStateMix
 				observable: observableTypes.action,
 				name: 'move-activity',
 				route: [route.activityUsage, route.collection]
+			},
+			classes: {
```
alexli01, line 51: Instead of relying on classes to determine if an item is a section, you should use "route:[route.activityUsage, route.collection]" to get the collection href, if the href exists, then it is a section.

alexli01 (approved):

PR #492, opened 2022-04-13
fix: learning path code rerendering upon input

tenbradley:
### Context
[DE48119](https://rally1.rallydev.com/#/?detail=/defect/631405651735&fdp=true)

The learning path code is rerendering and losing focus when updated. Switching over to lit's html function fixes this.


annabelsegalld2l (approved):

ChrisLabattD2L (approved):
Nothing was reloading/refreshing when I tested it, so hopefully this is the last one that needs to be changed like this! 😄

ChrisLabattD2L:
Side note, don't forget to add the `fix:` prefix to the merge this time 😜

PR #493, opened 2022-04-13
DE47656 [A11y] Does not give confirmation after moving questions

aid848:
[DE47656](https://rally1.rallydev.com/#/?detail=/defect/628618924251&fdp=true): [A11y] Does not give confirmation after moving questions– Hoping to include question name/title


mdulat:
I tested this out on my local, it seems to change the interaction when moving via keyboard quite a bit. I would demo this to Joseph first to see if it's ok. Also the focus is not being set when I move an item into a section the first time.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -412,8 +419,14 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 		await this._refreshNumberingState();
 
 		if (e.detail.keyboardActive) {
-			// US137145 Focus on item moved when keyboard active
-			this._toFocusKey = sourceItem.key;
```
mdulat, line 416: If we're not setting `_toFocusKey ` any more, do we need it at all? Here and also it's uses in `d2l-activity-collection-item-quiz.js`. (though not sure if we need it, given focus isn't being set when you move an item into a section now?)

aid848:
> I tested this out on my local, it seems to change the interaction when moving via keyboard quite a bit. I would demo this to Joseph first to see if it's ok. Also the focus is not being set when I move an item into a section the first time.

Yeah I'll run this by him to see if its ok like this.

components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
aid848: I don't think we need it as long as the focus is getting set consistently by the other method (which may not be the case in the move into section as you pointed out)

PR #499, opened 2022-04-19
fix: DE47656 announce items moved via keyboard up or down

mdulat:
Small part of: https://rally1.rallydev.com/#/?detail=/defect/628618924251&fdp=true

This only fixes the lack of announcement when an item is moved up in the list. 
This doesn't fix the focus issue at all.
It also doesn't announce what was moved, only the position.


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -662,6 +667,19 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
...
+		const list = this.shadowRoot?.querySelector('#quiz-object-list');
+
+		const newItem = list?.getListItemByKey(sourceKey);
+		newItem?.requestUpdate();
+		await newItem?.updateComplete;
+
+		const listItemDragHandle = newItem?.shadowRoot?.querySelector('d2l-list-item-drag-handle');
+		listItemDragHandle?.requestUpdate();
+		await listItemDragHandle?.updateComplete;
+		listItemDragHandle?.dispatchEvent(new CustomEvent('focusin', {bubbles: true, composed: true}));
```
mdulat, line 680: In d2l-list, the new label (i.e. announcement text of current (new) position) is constructed when the list-item-drag-handle gets the `focusin` event, so this dispatches that event whenever a move is made, to force it to create the proper announcement text.

aid848 (approved):

PR #502, opened 2022-04-22
fix: announce what was moved and keep focus

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/628618924251&fdp=true


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -297,6 +301,8 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 			<d2l-alert-toast type="success">
 				${this._successMessage}
 			</d2l-alert-toast>
+			<d2l-offscreen aria-live="assertive" aria-relevant="all">${this._moveByKeyboardConfirm}</d2l-offscreen>
+			<div class="d2l-offscreen" id="move_question_text" tabindex="-1"></div>
```
aid848, line 305: The offscreen element here that holds focus during the move seems to be generating the "blank" readout during the question move. I was hacking it with "..." to read nothing out earlier but I'm not sure if some text like "moving question" is better than blank or nothing being read out?

mdulat: ah good suggestion. I think before it had `"."` which made it read out `dot`. But I'll see if some other options would be better. The "blank" does get annoying to hear.

aid848 (approved):

PR #509, opened 2022-04-27
DE48184, DE47481: Show multiple pages of upcoming items if there are no overdue

mbardakovD2L:
Rally: [DE48184](https://rally1.rallydev.com/#/?detail=/defect/631795934473&fdp=true), [DE47481](https://rally1.rallydev.com/#/?detail=/defect/627952621395&fdp=true)


features/workToDo/d2l-w2d-collections.js
```diff
@@ -417,7 +417,7 @@ class W2dCollections extends LocalizeDynamicMixin(HypermediaStateMixin(LitElemen
 	}
 
 	get _pageUpcoming() {
-		if (!this._isOverduePastLastPage()) return 1;
```
mbardakovD2L, line 420: The root of the issue was here - `_isOverduePastLastPage()` always returns false if there are no overdue items, so `_pageUpcoming()` was always returning 1

features/workToDo/d2l-w2d-list.js
```diff
@@ -46,7 +47,7 @@ class W2dList extends HypermediaStateMixin(LitElement) {
 		return html`
 			<d2l-list separators="${this.collapsed ? 'none' : 'all'}">
 				${activities.map(activity => html`
```
mbardakovD2L, line 49: I tried changing this `map` to a `repeat`, but that just exacerbated the original issue, in that `w2d-d2l-list-item`s would just never render their contents at all

mbardakovD2L: I also tried changing the `html`s there to `lithtml`s, as per Brad's suggestion, but that resulted in exactly the same behaviour as the above comment

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -134,9 +136,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 		const oldHref = this.actionHref;
 		this._actionHref = href;
 		this.allowUnclickableActivities = true;
-		if (this.actionHref) {
-			this.requestUpdate('actionHref', oldHref);
-		}
```
mbardakovD2L, line 139: this was inconsistent with other places we do `set new value, requestUpdate with oldValue` - i.e. the other places we do this always do the `requestUpdate` without an if statement

features/workToDo/d2l-w2d-list-item.js
```diff
@@ -44,6 +44,7 @@ class W2DListItemMixin extends HypermediaStateMixin(ListItemLinkMixin(LocalizeDy
 			skeleton: { type: Boolean },
 			allowUnclickableActivities: { type: Boolean, attribute: 'allow-unclickable-activities' },
 			clickableFutureActivities: { type: Boolean, attribute: 'clickable-future-activities' },
+			_actionHref: { type: String },
```
mbardakovD2L, line 47: I couldn't see anywhere (in a parent class, mixin, or otherwise) where `_actionHref` was declared, despite being used - I think adding it back as a property is what ultimately led to the bug being fixed

features/workToDo/d2l-w2d-list-item.js
mbardakovD2L: on second thought, if we're trying to avoid extra re-renders, it might make sense to put this behind an if statement again now that I've fixed other problems with `actionHref`

features/workToDo/d2l-w2d-list-item.js
lloydwaddell: my concern with adding this here is that it's coming from the `foundations-engine`. I don't see anything that this property is being set either.

lloydwaddell: after the fix to `_isOverduePastLastPage`, does the issue still occur without this?

mbardakovD2L: after the fix to `_isOverduePastLastPage`, we can page to the different pages without getting the first page duplicate again, BUT the very first time you go to a page that's not the first page, the list item elements don't render properly - you have to page 1, then back to the page you were just looking at to get them to appear. if I declare `_actionHref` here, then list item elements are rendered successfully even on the first load

tenbradley: Looking in the foundation-engine, I don't see `_actionHref` being used anywhere, so this shouldn't cause an issue with the engine.

That being said, I'm surprised you it's being added as a property. I assume just adding `_actionHref` in the constructor isn't working?

features/workToDo/d2l-w2d-list-item.js
tenbradley: Are you having re-render issues?

mbardakovD2L: not re-render issues, but first-time-render issues - if I navigate away from page 2, then back to page 2 (for example), they render fine the second time

tenbradley (approved):

PR #511, opened 2022-04-29
US138635 Remove use of HMC render engine for rendering collection items.

mdgbayly:
Due to janky behaviour since Lit2 upgrade. Revert to using standard Lit template selection using a choose directive.

Since the recent Lit2 upgrade, we've noticed that rendering the FACE quiz collection list following reordering/dragndrop type interactions has become much more janky than previously. This is on top of the jankiness caused by the fact that we have to wait for the collection to be PATCHed and refetched from the server prior to re-rendering the changes.

It seems like the HMC rendered dynamic components are being recreated multiple times during re-rendering. This has always been problematic, and resulted in having to add all the additional render `guard` directives around the render templates.  But they seem to be less effective now. 

Given that the future of HMC is being questioned with the approaches being taken for UCE and given that in our case it is simple to replace the HMC render engine with more traditional programmatic Lit template selection (we only have 4 collection item types), we decided to just discontinue use of the HMC render engine for this feature.


alexli01 (approved):
Thanks Martin for fixing the problem.

mdulat (approved):

alexli01 (approved):

PR #518, opened 2022-05-06
US136655 Add pre-selected filter values

ChrisLabattD2L:
Context: [US136655](https://rally1.rallydev.com/#/?detail=/userstory/627979208241&fdp=true): [Discovery FRA] Add filter controller component to search page

Because of how the search page is being set up, we need a way to provide the filter with values that should already be selected when navigating to the page because the url will be keeping track of the filters that are currently active.


components/common/d2l-hc-filter-controller.js
```diff
@@ -21,20 +21,30 @@ class FilterController extends HypermediaStateMixin(LitElement) {
 		return html`
 			${repeat(this.filters, filter => html`
 				<d2l-hc-filter @d2l-filter-update="${this.selectedFiltersUpdated}"
+					id="${filter.properties.title}"
```
tenbradley, line 24: If we need an id, we could possibly use the orgUnitTypeId. We're not returning it from the bff currently, but we could.

components/common/d2l-hc-filter-controller.js
```diff
@@ -21,20 +21,30 @@ class FilterController extends HypermediaStateMixin(LitElement) {
...
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (changedProperties.has('filters')) {
+			this.initializeValues();
+		}
+	}
+
+	initializeValues() {
 		if (this.values.length === 0) {
```
tenbradley, line 40: nit: we might as well move this into the updated functions if condition?

components/common/d2l-hc-filter-controller.js
```diff
@@ -103,6 +113,29 @@ class FilterController extends HypermediaStateMixin(LitElement) {
 		this.requestUpdate();
 	}
 
+	updateSelectedValues(newFilterValues) {
+		newFilterValues.forEach(nfv => {
+			this.filters.forEach(filter => {
+				if (filter.properties.values.some(value => value.key.toString() === nfv)) {
```
tenbradley, line 119: This works, but I feel it could definitely be more efficient. It has me tempted to suggest we change the values field in the filters to be a set or map instead so we have quick lookup. This would prevent us from having to search the values for a matching key.

I just don't like how many nested for loops we have here. Maybe a function in the filter itself that takes `newFilterValues` instead so it can update itself?

components/common/d2l-hc-filter-controller.js
ChrisLabattD2L: That would make sense since the ID should be unique and there's a chance for the title to not be unique.

components/common/d2l-hc-filter-controller.js
ChrisLabattD2L: The reason I left it down here was because the `updated` function should typically be a simple function that calls other functions, I was just trying to keep that consistent with how we used the `updated` function in other files of ours. But this change can easily be made

components/common/d2l-hc-filter-controller.js
```diff
@@ -103,6 +113,29 @@ class FilterController extends HypermediaStateMixin(LitElement) {
 		this.requestUpdate();
 	}
 
+	updateSelectedValues(newFilterValues) {
```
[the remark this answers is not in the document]
ChrisLabattD2L: You make a fair point, I'll go ahead and change that!

components/common/d2l-hc-filter-controller.js
```diff
@@ -103,6 +113,29 @@ class FilterController extends HypermediaStateMixin(LitElement) {
 		this.requestUpdate();
 	}
 
+	updateSelectedValues(selectedValues) {
+		selectedValues.forEach(selectedValue => {
+			this.filters.forEach(filter => {
+				if (filter.properties.values.some(value => value.key.toString() === selectedValue)) {
```
tenbradley, line 119: As much as we are using keys that are numbers (orgUnitIds), this component is setup to have keys be strings. (At least I believe that is how I set it up.) In theory, we shouldn't have to convert the key to be a string since it should be a string, but I suppose doing this allows the keys to be numbers. If this is working and you're using the orgUnitIds as numbers (which is what the bff returns), then we can leave this to allow keys to be numbers.

components/common/d2l-hc-filter-controller.js
```diff
@@ -13,47 +13,54 @@ class FilterController extends HypermediaStateMixin(LitElement) {
...
+				this.initializeValues();
+			}
+		}
+	}
+
+	initializeValues() {
+		if (!Object.keys(this.values).length) {
 			this.filters.forEach(dimension => {
-				this.values.push({dimension: dimension.properties.title, selected: new Set()});
+				this.values[dimension.properties.title] = new Set();
```
BhamelinD2L, line 44: Should we be using the ID here to define sets? I think one of the concerns was something to do with two filters having the same name. Is there a potential issue here?

components/common/d2l-hc-filter-controller.js
```diff
@@ -13,47 +13,54 @@ class FilterController extends HypermediaStateMixin(LitElement) {
 
 	constructor() {
 		super();
-		this.values = [];
+		this.values = {};
```
BhamelinD2L, line 16: Coming into this blind, I'm not sure what this list is supposed to be. We're also sending `.values='${filter.properties.values}'` but that's different from this. It seems to be a set of filter names?

In the same way our rule picker has `rules, ruleIndex, conditions,` could we have more descriptive variables to explain what these things are?

Nothing appears to set this directly. Should it be underscored?

BhamelinD2L:
PR seems fine, just some small questions and I'm happy to approve once addressed

components/common/d2l-hc-filter-controller.js
ChrisLabattD2L: Yeah, you're 100% right, might as well go the whole way with using ID since the title isn't necessarily unique, didn't think of changing this when I changed the rest, I'll make the change!

components/common/d2l-hc-filter-controller.js
ChrisLabattD2L: So `this.values` is the variable that's assigned to keep track of which values are currently selected, meanwhile `.values='${filter.properties.values}'` is strictly referring to the possible values in the filters.

You do bring up a good point, it might be worthwhile to maybe change `this.values` to `this.selectedValues` so that it's more clear. I think the reason it's not underscored was so that others using this component would also be able to retrieve the selected values through it.

PR #521, opened 2022-05-10
fix: DE47968 Wrap offscreen elements in relatively positioned div to fix overflow issues

mdgbayly:
The offscreen elements that have been wrapped in a div were causing overflow because they were positioned relative to the document body because there were no other non-static elements in the component tree. So as the size of the question list got bigger this caused the overflow to increase. Even though it was technically hidden by css on the `primary-secondary-template`, it seemed to cause issues on FF and mobile browsers when we scrolled items into view using the [list-item-mixin's scrolling support](https://github.com/BrightspaceUI/core/blob/cbcae8a7acb7aa50a1e2c6c76d48bbf5ade63efe/components/list/list-item-mixin.js#L439). It seemed like the browser was scrolling the overflow into view.

Simply toggling the [`primary-secondary-template`'s `top` property](https://github.com/BrightspaceUI/core/blob/cbcae8a7acb7aa50a1e2c6c76d48bbf5ade63efe/templates/primary-secondary/primary-secondary.js#L520) on and off would cause the browser to recalculate and hide the overflow again, so seems like some weird browser bug.


components/activity/editor/collection/custom/quiz/d2l-activity-collection-editor-quiz.js
```diff
@@ -310,9 +313,11 @@ class ActivityCollectionEditorQuiz extends SkeletonMixin(HypermediaStateMixin(Lo
 			<d2l-alert-toast type="success">
 				${this._successMessage}
 			</d2l-alert-toast>
-			<d2l-activity-collection-item-lor-publish-error-dialog ?dialogOpened="${this._hasQpSelected}"/>
-			<d2l-offscreen aria-live="assertive" aria-relevant="all">${this._moveByKeyboardConfirm}</d2l-offscreen>
-			<div class="d2l-offscreen" id="move_question_text" aria-label="${this._movingByKeyboardAnnounce}" tabindex="-1"></div>
+			<d2l-activity-collection-item-lor-publish-error-dialog ?dialogOpened="${this._hasQpSelected}"></d2l-activity-collection-item-lor-publish-error-dialog>
```
mdgbayly, line 316: I also explicitly declared the end tag for this dialog element. I'm not sure if web components support self-closing tags like that. It had the bizarre side effect of adding any elements below (e.g. those offline elements) to the local DOM of unclosed dialog element. This was actually hiding the issue on FF and Mac Chrome/Safari, because it seemed to change the positioning of the offscreen elements, but it was still present on mobile browsers.

mdulat (approved):

PR #523, opened 2022-05-18
DE48362 Add fr-fr locale for list component

dustxd:
### Motivation/Context
* When the user account settings is set to `fr-fr`, the `fr` lang term is being used
  * `modifier la réserve de questions` (`fr-fr`)
  * `modifier le bassin de questions` (`fr`)
* `fr-fr` locale is missing in the localization file for list components which is why the offscreen component (i.e. accessibility label that is not visible on screen) is using the `fr` lang term

### Summary of Changes
Added `fr-fr` locale to the localization file for list components

### Rally Defect
[DE48362](https://rally1.rallydev.com/#/?detail=/defect/634194524971&fdp=true): [Face Quiz] > grades confirmation dialog lang terms not being picked up


mostlyfabulous (approved):

mostlyfabulous:
Oh there's manual import!?

dustxd:
> Oh there's manual import!?

I think lang files are still updated by the bot but to include the lang files in the localization of the component yes, it needs to be done manually lol each component does it in a different way too

PR #524, opened 2022-05-18
fix: DE48629 Attempt to minimize question list jumpiness when using k…

mdgbayly:
…eyboard to move items by moving focus to item specific offscreen elements based on the context of the move operation, rather than a single offscreen element.

* Added new offscreen elements at the top and bottom of each item item which can be used to retain focus and hopefully minize to much jumpiness. These elements are relatively positioned within the item to avoid overflow that was contributing to the footer overflow issues.
* Uses the source and target items and the context of the move operation to set focus to the item that we believe will not be disconnected by Lit element's repeat directive.
* Removed the existing offscreen element that was absolutely positioned.
* Removed the lang terms associated with the existing offscreen element. We can't use an announcement now when we are moving focus as the elements are now per item and this was also being announced when the item was focused using standard list navigation arrow keys.


mdulat (approved):
Let's give it a try!

aid848 (approved):

mdgbayly:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F637839981207

PR #526, opened 2022-05-18
US138981 auto-zero grade annotate component - US138981

imcdonald-d2l:
auto-zero-annotate-component-demo

Two ways to enable annotate:

1. Add HM entity `href` with appropriate `isAutoZeroed` property. (haven't tested this in a "real" environment - I assume this works based off other foundation component setups)
2. Add `force-show-auto-zero` attribute and it will annotate in the case where we determine this from the back-end and don't need to fetch an entity from the component.

BSI PR: https://github.com/Brightspace/brightspace-integration/pull/6456

[Rally ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F635458885621)


DiljotSG (approved):
LGTM

PR #536, opened 2022-05-30
Eandrews/us139435

emsandrews:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F638022661061

Update styling for External Activity Question list item in FACE Quiz. 

1. update Activity name text to `d2l-body-compact` in `d2l-activity-list-item-activity` component. 
2. update `d2l-hc-name component` and `d2l-activity-name-specialization` so activity question text wraps when window is resized.


components/activity/name/custom/d2l-activity-name-specialization.js
```diff
@@ -19,7 +19,7 @@ class ActivityNameSpecialization extends LabelMixin(HypermediaStateMixin(LitElem
 
 	render() {
 		return html`
-			<d2l-hc-name href="${ifDefined(this._specializationHref)}" .token="${this.token}"></d2l-hc-name>
+			<d2l-hc-name wrap href="${ifDefined(this._specializationHref)}" .token="${this.token}"></d2l-hc-name>
```
mdgbayly, line 22: Suggestion: Might want to expose this `wrap` as a property of `d2l-activity-name-specialization` too if we are concerned about affecting other usages, and then setting it from `d2l-activity-list-item-activity`.

mdgbayly (approved):
LGTM

PR #537, opened 2022-05-31
Filtered values can be unselectable

tenbradley:
### Context
[US137139](https://rally1.rallydev.com/#/?detail=/userstory/629722256133&fdp=true)
Since the core ui filter component has been updated to allow values to be disabled, we are updating our use of the filters to disable filters instead of hiding them when the values are updated and old values are not present.


components/common/d2l-hc-filter.js
```diff
@@ -17,42 +17,54 @@ class HmFilter extends LitElement {
...
 	render() {
 		if (!this.title || !this._values || !this.key) return;
 		return html`
 			<d2l-filter @d2l-filter-change="${this._onFilterChange}">
-				<d2l-filter-dimension-set key="${this.key}" text=${this.title}>
-					${repeat(this._values, (value) => value.key, value => html`
-						<d2l-filter-dimension-set-value key="${value.key}" text="${value.text}" ?selected="${this.selected.get(value.key)}"></d2l-filter-dimension-set-value>
-					`)}
+				<d2l-filter-dimension-set key="${this.title}" text=${this.title}>
+				${repeat(this.data, (value) => value[1].selected, value => html`
```
svanherk, line 28: So I don't know if this will affect the problem you're seeing, but I expect you want this to actually be: 
```suggestion
				${repeat(this.data, (value) => value[0], value => html`
```
That second parameter is supposed to be a function that returns a unique id, so that Lit knows not to reuse the dom nodes if one is added or removed unless it has the same id

tenbradley: Makes sense. I wasn't quite sure why this was needed.

I tried that locally and it's behaving the same way. Is there something I should be doing differently with how I'm interacting with the filter?

